PR #9, opened 2016-02-15
Fix webview install on ubuntu 14.04

karenc:
It was failing on step "install nodejs and npm packages (on Debian)"
because /usr/local/share/systemtap/tapset doesn't exist on ubuntu 14.04


pumazi:
Merged.

PR #10, opened 2016-02-17
Include script to create rhaptos site

karenc:
Also have a look at the changes I made to the create_rhaptos_site.py script:

https://github.com/karenc/cnx-buildout/commit/8689ba75a7a96afa02631ccc8fb3a5f220ca5553

`manage_importSelectedSteps` seems to require an empty database (https://github.com/Rhaptos/Products.RhaptosModuleStorage/blob/master/Products/RhaptosModuleStorage/setuphandlers.py#L260), I'm not sure how it's supposed to work.

### DIFF
--- a/tasks/install_cnx_buildout.yml
+++ b/tasks/install_cnx_buildout.yml
@@ -110,5 +110,22 @@
 - name: start zeoserver
   shell: chdir="{{ source_dir }}/cnx-buildout" ./bin/zeoserver start
 
+# +++
+# Create Rhaptos Site
+# +++
+
+- name: check whether rhaptos site exists
+  register: rhaptos_site
+  shell: echo 'print(app.site)' | ./bin/instance debug | grep PloneSite
+  args:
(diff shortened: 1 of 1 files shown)


karenc:
@reedstrm Could you have a look at this?

pumazi:
I have verified this does indeed create a rhaptos site. I haven't yet verified that the data is pulled in.

karenc:
I don't know how the data is supposed to be pulled in.  @reedstrm ?

pumazi:
Yes, right. I used the wrong words. I changed it right after, but I guess I was not fast enough for it to make it into the email.

pumazi:
This doesn't seem to assign the correct host for the postgresql connection, but this is probably a connection problem on my part rather than the change itself, because it's using the value of archive_db_host.

I'll merge it shortly. :+1:

pumazi:
Merged.

PR #14, opened 2016-02-19
Allow different users to deploy to the same host

karenc:
- :bug: Fix permissions for planted source
 - Build webview on debian without using sudo
 - Install webview as www-data on debian systems
 - Install python packages as www-data on debian systems

### DIFF
--- a/tasks/install_arclishing.yml
+++ b/tasks/install_arclishing.yml
@@ -72,15 +72,28 @@
     path: "{{ root_prefix }}/var/cnx/venvs"
     state: directory
     mode: 0755
-    owner: "{{ ansible_user_id }}"
+    owner: "{{ ansible_os_family == 'Darwin' and ansible_user_id or 'www-data' }}"
+
+- name: set the owner of venvs directory (on Debian)
+  when: ansible_os_family == 'Debian'
+  become: yes
+  file:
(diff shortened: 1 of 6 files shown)


tasks/install_arclishing.yml
```diff
@@ -72,15 +72,28 @@
...
     mode: 0755
-    owner: "{{ ansible_user_id }}"
+    owner: "{{ ansible_os_family == 'Darwin' and ansible_user_id or 'www-data' }}"
+
+- name: set the owner of venvs directory (on Debian)
+  when: ansible_os_family == 'Debian'
+  become: yes
+  file:
+    path: "{{ root_prefix }}/var/cnx/venvs"
+    state: directory
```
pumazi: I'm wondering if this task should be using 'touch' rather than directory. Or whether it makes a difference.

pumazi:
I see the ``webview/dist`` directory is still assigned to the ansible_user_id. We may need to touch up that task as well.

karenc:
Since `webview/dist` is on the local machine, I think it is ok to leave it assigned to the ansible_user_id.

karenc:
Ok I think I got everything that I want in.  Ready for review.

PR #16, opened 2016-02-24
deployment fixes

aw30:
check changes, let me know what may need to be changed/updated.


karenc:
I think most of the modified files shouldn't be committed (e.g. commenting out code).

pumazi:
I've added 8c906c0b1c7efc86e5deef81f016d4aadc95940d, which commits the qa environment.

The changes to the playbook have not been committed. The playbook changes disable git checkouts and remove varnish. Varnish is specifically important because it links all the serves together under one domain.

We will possibly have a different solution for the git issues experienced here (i.e. timeouts). We are looking into potentially using [manifestly](https://github.com/openstax/manifestly) for checkouts and versioning.

PR #19, opened 2016-03-03
Update acquire_data.yml after moving columns in archive files tables

karenc:
This is for after the database is migrated. (I tested this on cte-cnx-dev)


pumazi:
Aha, this explains why it stopped working. :+1:

pumazi:
Does this mean the media-type is lost in this procedure?

karenc:
We have `SELECT * FROM files` so it's actually ok.

PR #20, opened 2016-03-04
Remove dummy file insertion in acquire_data and push_data

karenc:
Instead, just change all the file content (column "file" in files table)
to "dummy file".

This allows us to keep all the entries in the files table, with all the
md5 and sha1 hashes.


pumazi:
Followup enhancement at #22.

PR #24, opened 2016-03-15
Remove plpydbapi from archive and publishing

karenc:


pumazi:
Very nice! And fast. :stuck_out_tongue:

I'll get it in as soon as I've merged Connexions/cnx-archive#399

Issue #30, opened 2016-04-06
Rhaptos Site creation fails when database already exists


pumazi:
## Summary

The rhaptos site creation script in `tasks/install_cnx_buildout.yml` fails when the database is already initialized. Furthermore, it doesn't error and returns status code 0, which makes it look successful.

## Environment

cte-dev on Ubuntu 14.04

pumazi:
Because of this, the first run needs to be against a blank database. Then one must sign into the ZMI and change the database connection info within the plone site.

PR #33, opened 2016-04-13
plone.yml fixes

karenc:
:skull: Don't use the merge button.


pumazi:
I was kind of hoping we would fix this issue within the create_rhaptos_site script in cnx-buildout. This work around should be fine though.

This means we'll continue to have an empty postgresql service running on each deploy of zope. :confounded:

karenc:
Oh... :stuck_out_tongue: you're right
I'll try doing it in the script.

pumazi:
Closes #30

PR #42, opened 2016-04-28
Fix provision accounts playbook

karenc:
Together with https://github.com/openstax/tutor-deployment/pull/234


reedstrm:
tested locally, works for me.

PR #62, opened 2016-07-30
Populate the ZODB from the postgres database entries

### LINKED ISSUE #47: Populate the ZODB from the database entries
## Summary

Populate the ZODB with object references to the content in the SQL database. One way to do this is to hit every piece of content's url. This however loads the data on each request, which isn't something we really care about. 

@reedstrm has successfully done this using the following steps:
```
psql -l -U postgres repository -t -c "select 'http://legacy-cte-cnx-dev.cnx.org/content/'||moduleid||'/latest' from latest_modules order by portal_type,revised" > legacy_urls.txt
wget -q -i legacy_urls.txt
```

An alternative and more efficient means to do this is to develop a zope script to do this. The script would loop through the modules in the database. In the body of the loop we would "get-by-id" for each object. Then would effectively create the object in the ZODB without loading the entire contents.

This work wouldn't necessarily happen in cnx-deploy, but it would primarily be used here. The likely location of this script would be in cnx-buildout.

## Environment

all

karenc:
When a user views a page on zope, if the object doesn't exist in the
ZODB but is in the postgres database, there is code to create the ZODB
objects.  This commit gets all the urls for collections and modules for
the zope site and use wget to access them so the ZODB objects are
created.

Close #47

Set `archive_db_host` to `cte-cnx-dev.cnx.org` for `cte-legacy-dev.cnx.org`

---

:skull: **DO NOT MERGE this pull request** :skull: ￼￼

### DIFF
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -39,6 +39,14 @@
     name: rhaptos
     state: absent
 
+- name: create zope objects from database (background job)
+  when: rhaptos_site|failed
+  become: yes
+  become_user: www-data
+  shell: ./bin/instance run scripts/build_zodb_from_postgres.py
+  args:
+    chdir: "{{ source_dir }}/cnx-buildout"
+
 - name: start instance
   become: yes
   shell: supervisorctl restart plone-instance
--- a/environments/cte-dev/host_vars/cte-legacy-dev.cnx.org
+++ b/environments/cte-dev/host_vars/cte-legacy-dev.cnx.org
@@ -1,3 +1,5 @@
 ---
 source_dir: /var/lib/cnx
 local_source_dir: "{{ '~/cnx/src' | expanduser }}"
+
+archive_db_host: cte-cnx-dev.cnx.org


tasks/create_rhaptos_site.yml
```diff
@@ -42,3 +42,28 @@
 - name: start instance
   become: yes
   shell: supervisorctl restart plone-instance
+
+- name: send the script to create list of zope urls
+  when: rhaptos_site|failed
+  become: yes
+  template:
+    src: "misc/zope_urls_from_db.py"
```
pumazi: Any reason we shouldn't put this script in the cnx-buildout repo? Seems to me to be more aligned with that repo's purpose and existence.

templates/misc/zope_urls_from_db.py
```diff
@@ -0,0 +1,22 @@
+import psycopg2
+
+
+connection_string = (
+    'postgres://{{ archive_db_user }}:{{ archive_db_password }}'
```
pumazi: Probably not a great idea to hardcode the database password in a script. The file is readily accessible to anyone that can user authenticate. Basically, it means a backdoor hack is possible, which could easily be used to compromise the whole machine.

tasks/create_rhaptos_site.yml
karenc: When I was writing this, I thought, this is only necessary because we have an existing database, so it shouldn't be in cnx-buildout. But yes, I think it makes sense to have it in cnx-buildout. I'll also try to redo the script so it doesn't use wget.

templates/misc/zope_urls_from_db.py
karenc: Yes, we actually have the database details in zope so I can just use `./bin/instance run`.

karenc:
Ready for review again, see also https://github.com/Rhaptos/cnx-buildout/pull/11

pumazi:
If I'm not mistaken the first commit isn't really needed now.

This is much better! 💯 🎈

karenc:
The first commit is necessary, otherwise on cte-legacy-dev the db connection string is to localhost which is wrong.

PR #63, opened 2016-07-31
Run cnx-archive-init_venv in case venv is not set up in postgres

karenc:
Python virtualenv paths need to be loaded in postgres so it's possible
to use cnx packages in postgres.  cnx-archive-init_venv sets this up.

Depends on https://github.com/Connexions/cnx-archive/pull/448

---

:skull: **DO NOT MERGE this pull request** :skull:


karenc:
This was originally done (urgently) for cte-dev because venv was not working after copying the production database.  Going to redo this later with cnx-db.

PR #64, opened 2016-07-31
Set up stub accounts if accounts_domain is not set

karenc:
Set the openstax_accounts stub flag if accounts_domain is not set.  For
example using the vm environment, if accounts is not set up, it is
possible to use the stub by doing:

```
ansible-playbook -i environments/vm/inventory main.yml -e accounts_domain=
```

---

### DIFF
--- a/templates/etc/cnx/authoring/app.ini
+++ b/templates/etc/cnx/authoring/app.ini
@@ -28,6 +28,7 @@ current-license-urls =
 # size limit of file upload in MB
 authoring.file_upload.limit = 50
 
+{% if accounts_domain %}
 {% if (accounts_disable_verify_ssl is defined and accounts_disable_verify_ssl)
       or inventory_dir.endswith('environments/local') %}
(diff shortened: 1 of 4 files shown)


pumazi:
In development this makes sense, but I don't think it should be the default on a deployment, which it would be since most people won't know to set it.

I'm feeling like this adds weedy bits to our templates. Especially because it's only checking for a null configuration option.

Personally I like the use of the stub feature in development. However, I feel indifferent about this change. I don't think we have yet completed the full scenario that works towards making a rock solid **production** deploy. So, adding the development scenario in at this time seems premature.

karenc:
I wouldn't mind not having it in cnx-deploy if there's just one config option to turn on stub accounts on the vm, but that's not the case.  If there's another way of doing this so it's possible to merge to master, I would be happy to change it.

karenc:
We can just leave this branch around if someone needs to set up stub accounts.

PR #65, opened 2016-08-05
Fix the installation of cssselect2 with an additional argument to pip

pumazi:
The ``exists-action`` setting will ``(s)witch`` to the newly requested
install. Otherwise pip would prompt for user input and as such fail
in ansible.

This problem only occurs because we switched from using the parent
author's version of the project to our connexions' fork.

---

### DIFF
--- a/tasks/install_publishing.yml
+++ b/tasks/install_publishing.yml
@@ -180,6 +180,7 @@
     name: "git+https://github.com/Connexions/cssselect2.git#egg=cssselect2"
     virtualenv: "{{ root_prefix }}/var/cnx/venvs/publishing"
     state: present
+    extra_args: "--exists-action s"
 # /FIXME cssselect2-acquisition
 
 - name: install (local) publishing packages


reedstrm:
So, what's the process for these "DO NOT MERGE" repos? That mainly means "don't push the big green button", right? Got a gist of the expected process?

karenc:
We don't want to have merge commits, but we want to keep all the commits in the pull requests, so github merge button doesn't really do this.

1. Make sure your local repo is up to date:

    ```
    git remote update -p
    ```
2. Check out the branch you want to merge and make sure it's rebased:

    ```
    git checkout <branch-name>
    git rebase origin/master
    git push origin HEAD
    ```
3. Merge this branch to master:

    ```
    git checkout master
    git reset --hard origin/master
    git merge --ff-only <branch-name>  # I haven't actually tried --ff-only, but looks right
    git push origin HEAD
    ```

This is actually a standard way of merging code without merge commits, we didn't think it was necessary to include a gist.

reedstrm:
"That's the wonderful thing about standards, there are so many to choose from" :-) So basically, resolve all potential merge conflicts via rebasing feature branch, ff-merge only to master. Got it.

reedstrm:
Process docs are good - Ah, I see the push to origin after rebase is in there, so that'll trigger travis/PR tests.

reedstrm:
However, 'git push origin HEAD' is going to be a force update, in pretty much every case.

karenc:
The first `git push origin HEAD` (the PR branch) sometimes will be a force update but the second `git push origin HEAD` to master shouldn't need a force update.

PR #67, opened 2016-08-10
Update data playbooks

pumazi:
- Wraps the commands in the acquisition playbook. This is to make it compatible with future ansible versions.
- Adds a task to create the database user during push. This helped in my local situation where I did not have a rhaptos user created before attempting to push the data onto the server.

---

### DIFF
--- a/push_data.yml
+++ b/push_data.yml
@@ -10,7 +10,10 @@
     - name: "db_user"
       prompt: "What database user should we use?"
       default: postgres
+    - name: "db_password"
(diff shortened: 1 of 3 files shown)


push_data.yml
```diff
@@ -10,7 +10,10 @@
     - name: "db_user"
       prompt: "What database user should we use?"
       default: postgres
+    - name: "db_password"
+      prompt: "What database user should we use?"
```
karenc: This should probably be `What password should we use for the database user?` or something, right?

PR #71, opened 2016-08-16
Setup for PDF generation within zope

reedstrm:
~~@karenc untested from ansible, but the shell command was tested on legacy-dev.~~ :+1:

### DIFF
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -27,6 +27,12 @@
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
+- name: set up localFS PDF folder
+  when: rhaptos_site|failed
+  shell: "./bin/instance run scripts/set_up_localfs_for_pdfs.py"
+  args:
+    chdir: "{{ source_dir }}/cnx-buildout"
+
 - name: update database connection string in rhaptos site
   when: rhaptos_site|failed
   shell: echo 'import transaction; app.plone.rhaptosDA.connection_string = "postgres://{{ archive_db_user }}:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ archive_db_name }}"; transaction.commit()' | ./bin/instance debug
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -7,23 +7,18 @@
     chdir: "{{ source_dir }}/cnx-buildout"
   ignore_errors: True
 
-- name: create rhaptos database user
-  when: rhaptos_site|failed
-  vars:
-    db_user: rhaptos
-    db_password: "{{ archive_db_password }}"
-  include: tasks/create_db_user.yml
-
 - name: create empty temporary rhaptos database
   when: rhaptos_site|failed
   vars:
     db_name: rhaptos
-    db_owner: rhaptos
+    db_owner: "{{ archive_db_user }}"
+    db_host: "{{ archive_db_host }}"
+    db_port: "{{ archive_db_port }}"
   include: tasks/create_db.yml
 
 - name: create rhaptos site
   when: rhaptos_site|failed
-  shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos localhost 5432"
+  shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos {{ archive_db_host }} {{ archive_db_port }}"
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
@@ -43,6 +38,8 @@
   when: rhaptos_site|failed
   postgresql_db:
     name: rhaptos
+    login_host: "{{ archive_db_host }}"
+    port: "{{ archive_db_port }}"
     state: absent
 
 - name: start instance
--- a/tasks/create_db.yml
+++ b/tasks/create_db.yml
@@ -10,6 +10,8 @@
 - name: create the archive db
   postgresql_db:
     login_user: postgres
+    login_host: "{{ db_host|default('localhost') }}"
+    port: "{{ db_port|default(5432) }}"
     name: "{{ db_name }}"
     encoding: "utf-8"
     # FIXME The owner should not be the same as the user.
--- a/tasks/create_db_user.yml
+++ b/tasks/create_db_user.yml
@@ -10,6 +10,8 @@
 - name: create the db user
   postgresql_user:
     db: postgres
+    login_host: "{{ db_host|default('localhost') }}"
+    port: "{{ db_port|default(5432) }}"
     login_user: postgres
     name: "{{ db_user }}"
     password: "{{ db_password }}"
--- a/tasks/install_cnx_buildout.yml
+++ b/tasks/install_cnx_buildout.yml
@@ -57,12 +57,13 @@
     - ruby
     - inkscape
     - docbook-xsl-ns
-    - openjdk-6-jre
+      #- openjdk-6-jre
     - texlive-fonts-recommended
     - memcached
     - librsvg2-bin
     - otf-stix
-    - openjdk-7-jdk
+      #- openjdk-7-jdk
+    - openjdk-8-jdk
 
 # +++
 # Install virtualenv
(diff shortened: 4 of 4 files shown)


tasks/create_rhaptos_site.yml
```diff
@@ -7,36 +7,58 @@
...
   when: rhaptos_site|failed
-  vars:
-    db_user: rhaptos
-    db_password: "{{ archive_db_password }}"
-  include: tasks/create_db_user.yml
+  postgresql_user:
+    name: rhaptos
+    password: "{{ archive_db_password }}"
+    login_host: "{{ archive_db_host }}"
+    port: "{{ archive_db_port }}"
```
pumazi: Wouldn't it be better to make the create_db_user task more robust rather than create a copy of it? Also, this doesn't ensure the prerequisite dependency (`psycopg2`) has been installed on the system.

tasks/create_rhaptos_site.yml
```diff
@@ -7,36 +7,58 @@
...
   when: rhaptos_site|failed
-  vars:
-    db_name: rhaptos
-    db_owner: rhaptos
-  include: tasks/create_db.yml
+  postgresql_db:
+    name: rhaptos
+    owner: rhaptos
+    login_host: "{{ archive_db_host }}"
+    port: "{{ archive_db_port }}"
```
pumazi: ditto last comment

tasks/create_rhaptos_site.yml
```diff
@@ -7,36 +7,58 @@
...
 
 - name: create rhaptos site
   when: rhaptos_site|failed
   shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos localhost 5432"
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
+- name: create localFS PDF folder
+  when: rhaptos_site|failed
+  shell: echo 'import transaction; app.plone._delObject("pdfs"); app.plone.manage_addProduct["LocalFS"].manage_addLocalFS(id="pdfs",title="pdf-storage",basepath="/var/www/files"); transaction.commit()' | ./bin/instance debug
```
pumazi: Maybe this should be a script in cnx-buildout?

tasks/create_rhaptos_site.yml
```diff
@@ -7,36 +7,58 @@
...
+
 - name: update database connection string in rhaptos site
   when: rhaptos_site|failed
   shell: echo 'import transaction; app.plone.rhaptosDA.connection_string = "postgres://{{ archive_db_user }}:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ archive_db_name }}"; transaction.commit()' | ./bin/instance debug
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
+- name: fix rhaptos print tool storage paths
+  when: rhaptos_site|failed
+  shell: echo 'import transaction; ptool = app.plone.rhaptos_print; ptool.storagePaths = ptool.DEFAULT_STORAGE_PATHS; ptool.storagePath = ptool.storagePaths[0]; transaction.commit()' | ./bin/instance debug
```
pumazi: ditto last comment

tasks/create_rhaptos_site.yml
karenc: You're right.  I'll change it.

tasks/create_rhaptos_site.yml
karenc: I've thought about it.  I'll move it to cnx-buildout.

karenc:
Add LocalFS using a script in cnx-buildout: https://github.com/Rhaptos/cnx-buildout/pull/17 :wink:

karenc:
Ready for review again

pumazi:
Looks 💯 ❗ 

I'll give it a try as soon as possible. 🎱

pumazi:
The first try seems to have produced this issue:

```
...
TASK [create rhaptos site] *****************************************************
changed: [rubble]

TASK [set up localFS PDF folder] ***********************************************
changed: [rubble]

TASK [update database connection string in rhaptos site] ***********************
changed: [rubble]

TASK [remove temporary rhaptos database] ***************************************
changed: [rubble]

TASK [remove temporary rhaptos database user] **********************************
fatal: [rubble]: FAILED! => {"changed": false, "failed": true, "msg": "unable to remove user"}
```

pumazi:
The playbook went through to completion on a followup run. This is likely because that failing task was skipped this time around. I'll give it a fresh run...

pumazi:
Unfortunately, I'm still seeing a failure during user removal.

I did resolve the issue by removing whole chunks of the task file. The patch is as follows:

```patch
diff --git a/tasks/create_rhaptos_site.yml b/tasks/create_rhaptos_site.yml
index 98d98a5..80046f6 100644
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -7,24 +7,6 @@
     chdir: "{{ source_dir }}/cnx-buildout"
   ignore_errors: True
 
-- name: create temporary rhaptos database user
-  when: rhaptos_site|failed
-  vars:
-    db_user: rhaptos
-    db_password: "{{ archive_db_password }}"
-    db_host: "{{ archive_db_host }}"
-    db_port: "{{ archive_db_port }}"
-  include: tasks/create_db_user.yml
-
-- name: create empty temporary rhaptos database
-  when: rhaptos_site|failed
-  vars:
-    db_name: rhaptos
-    db_owner: rhaptos
-    db_host: "{{ archive_db_host }}"
-    db_port: "{{ archive_db_port }}"
-  include: tasks/create_db.yml
-
 - name: create rhaptos site
   when: rhaptos_site|failed
   shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos localhost 5432"
@@ -43,22 +25,6 @@
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
-- name: remove temporary rhaptos database
-  when: rhaptos_site|failed
-  postgresql_db:
-    name: rhaptos
-    login_host: "{{ archive_db_host }}"
-    port: "{{ archive_db_port }}"
-    state: absent
-
-- name: remove temporary rhaptos database user
-  when: rhaptos_site|failed
-  postgresql_user:
-    name: rhaptos
-    login_host: "{{ archive_db_host }}"
-    port: "{{ archive_db_port }}"
-    state: absent
-
 - name: start instance
   become: yes
   shell: supervisorctl restart plone-instance
```

karenc:
I'm pretty sure the temporary database is necessary...

karenc:
Isn't it because the create_rhaptos_site.py fails if archive db is already populated?

karenc:
I made this change and it worked for me:

```diff
diff --git a/tasks/create_rhaptos_site.yml b/tasks/create_rhaptos_site.yml
index 98d98a5..a3dce6b 100644
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -27,7 +27,7 @@
 
 - name: create rhaptos site
   when: rhaptos_site|failed
-  shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos localhost 5432"
+  shell: "./bin/instance run scripts/create_rhaptos_site.py plone Portal postgres {{ archive_db_user }} rhaptos {{ archive_db_host }} {{ archive_db_port }}"
   args:
     chdir: "{{ source_dir }}/cnx-buildout"
 
```

pumazi:
I'm guessing this is because I've configured my environment to reflect production using the `rhaptos` database user. The problem task is trying to remove the user, but it owns another database. How about we call these `rhaptos_temp_db` and `rhaptos_temp_user`?

karenc:
Oh I see, yes I guess we could do that, the reason I called it `rhaptos` was so the path of the object is `app.plone.rhaptosDA` (`app.plone.{{ db_name }}DA`), but we can rename the object?

pumazi:
Ahh yeah, I just noticed the `rhaptosDA` object. Let's not rename it, since zope sometimes does weird things when renaming. Keep the db name. But let's use a different temp user.

karenc:
Ok, good idea, let me try that :)

pumazi:
The other thing... Why are these being created on the main database server rather than the localhost. I thought this was the primary reason we needed postgres installed on the host that contains zope?

karenc:
I was going to remove postgresql on zope hosts, so I started doing that and installing `libpq-dev`, but then I saw that the buildout config files actually use postgresql location...  I'm not sure if I can remove postgresql

pumazi:
There is the chance the site will fail to create. The last line of the traceback is `ValueError: Database populated, not using to avoid dataloss` In this case, the task does not fail. So it's possible to rerun this task several times without understanding what the problem is.

pumazi:
My adjustments to the most recent changes are to do the following:

```patch
diff --git a/tasks/create_rhaptos_site.yml b/tasks/create_rhaptos_site.yml
index a3dce6b..46cdd27 100644
--- a/tasks/create_rhaptos_site.yml
+++ b/tasks/create_rhaptos_site.yml
@@ -7,20 +7,11 @@
     chdir: "{{ source_dir }}/cnx-buildout"
   ignore_errors: True
 
-- name: create temporary rhaptos database user
-  when: rhaptos_site|failed
-  vars:
-    db_user: rhaptos
-    db_password: "{{ archive_db_password }}"
-    db_host: "{{ archive_db_host }}"
-    db_port: "{{ archive_db_port }}"
-  include: tasks/create_db_user.yml
-
 - name: create empty temporary rhaptos database
   when: rhaptos_site|failed
   vars:
     db_name: rhaptos
-    db_owner: rhaptos
+    db_owner: "{{ archive_db_user }}"
     db_host: "{{ archive_db_host }}"
     db_port: "{{ archive_db_port }}"
   include: tasks/create_db.yml
@@ -51,14 +42,6 @@
     port: "{{ archive_db_port }}"
     state: absent
 
-- name: remove temporary rhaptos database user
-  when: rhaptos_site|failed
-  postgresql_user:
-    name: rhaptos
-    login_host: "{{ archive_db_host }}"
-    port: "{{ archive_db_port }}"
-    state: absent
-
 - name: start instance
   become: yes
   shell: supervisorctl restart plone-instance
```

karenc:
Ok, that makes sense.  I've made the changes.

pumazi:
We'll likely need to revisit this again in the future when we've corrected #75, but until then, this should work. 👍

PR #81, opened 2016-08-25
Allow zope to connect to postgres and set correct archive db host for zope

karenc:


pumazi:
Does this actually work for you? In order to make this work I needed to use the specific hostname. But this was prior to the configure_hosts playbook. At that time I was doing it by hand and it wasn't working. Because of this I figured we would make best practice for the VM scenario to put everything on one VM.

karenc:
It works for me, but I haven't tried it from scratch.

Oh does configure_hosts playbook not work for you?  It should set up the hosts correctly according to your local /etc/hosts.

pumazi:
The need for this has been removed in #84.

PR #91, opened 2016-09-02
Install and configure HAProxy to proxy multiple Zope instances

pumazi:
This configures, builds and adds init configuration for multiple client instances.
This can be configured using the host variable `zclient_count`.

There is a minor configuration filename change here. `/etc/supervisor/conf.d/zclient.conf` was changed to `/etc/supervisor/conf.d/zclients.conf`(plural). And this configuration does not collide with the previous one so it's not exactly harmful, other than the main `instance` is started, when it shouldn't be. Removing the file corrects the issue. Since we haven't doing any push of this anywhere recently, I don't see a need to put in a task that removes the offending file.

### DIFF
--- a/environments/cte-dev/group_vars/all/vars.yml
+++ b/environments/cte-dev/group_vars/all/vars.yml
@@ -21,8 +21,6 @@ memcached_hosts: localhost
 
 varnish_port: 80
 
-zope_host: cte-legacy-dev.cnx.org
-zope_port: 8888
 archive_host: archive-cte-cnx-dev.cnx.org
 archive_port: 6789
 publishing_host: archive-cte-cnx-dev.cnx.org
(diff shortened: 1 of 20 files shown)


pumazi:
The last commit is basically reverted in a followup branch for HAProxy. I may combine that branch into this one and drop the commit.

karenc:
`haproxy` is not defined.  Is this the right thing to do?

```diff
diff --git a/environments/vm/inventory b/environments/vm/inventory
index b823812..817531a 100644
--- a/environments/vm/inventory
+++ b/environments/vm/inventory
@@ -39,6 +39,9 @@ zope.local.cnx.org
 [pdf_gen:children]
 zope.local.cnx.org
 
+[haproxy:children]
+local.cnx.org
+
 # Groups of groups
 
 [db_connected:children]
```

pumazi:
@karenc Yes, that is correct. I've revised the commit.

karenc:
:100:

PR #92, opened 2016-09-02
Add nfs directories to archive configuration file

karenc:


templates/etc/cnx/archive/app.ini
```diff
@@ -38,6 +38,8 @@ search-long-cache-expiration = 43200
 
 exports-directories =
     /usr/local/var/cnx/archive/exports
```
pumazi: Is this directory still needed?

karenc: I guess not :p

karenc: Removed

PR #96, opened 2016-09-07
Add accounts to /etc/hosts when running configure_hosts.yml

karenc:


configure_hosts.yml
```diff
@@ -16,7 +16,7 @@
     - pdf_gen
   tasks:
     - name: get dns from local /etc/hosts
-      local_action: "command grep -E '^${% for group in groups.all %}|{{ '\\<%s\\>'|format(group) }}{% endfor %}' /etc/hosts"
+      local_action: "command grep -E '\\<{{ accounts_domain }}\\>{% for group in groups.all %}|{{ '\\<%s\\>'|format(group) }}{% endfor %}' /etc/hosts"
```
pumazi: Can you explain what this is doing?

karenc: It used to be:

```
grep -E "^$|\<local.cnx.org\>|\<zope.local.cnx.org\>|\<127.0.0.1\>" /etc/hosts
```

It is now:

```
grep -E "\<accounts.local.openstax.org\>|\<local.cnx.org\>|\<zope.local.cnx.org\>|\<127.0.0.1\>" /etc/hosts
```

getting the `/etc/hosts` entries for `local.cnx.org` (like in python `\blocal.cnx.org\b`) and `zope.local.cnx.org`.  Then I realized it's not looking for `accounts.local.openstax.org` which I needed in the VM.  Does that answer your question?

pumazi: Yes, thanks.

 I'm guessing grep was likely easier to debug than the `lineinfile` module. So, good choice.

karenc: This is actually the only way I know how to get the `/etc/hosts` entries from localhost :sweat_smile:

pumazi: O... actually, I misunderstood anyway. I get it now. 👍

PR #100, opened 2016-09-21
Creates ansible roles for the front-end components

pumazi:
This is a rough stab at creating roles for the front-end components of CNX.


pumazi:
@dtwilliamson @DarkPrinceFrost Thoughts and opinions?

PR #103, opened 2016-09-29
Download and unpack prebuilt webview from packages.cnx.org

pumazi:
This will download the prebuilt version of webview from
https://packages.cnx.org/js-builds/webview_$version.tar.gz
The contents are unpacked and moved into /var/www and the
new version is linked into place.


pumazi:
This is currently blocked by the acceptance of Connexions/rundeck-scripts#1. Once that is deployed this will be functional.

roles/webview/tasks/webview.yml
```diff
@@ -5,24 +5,81 @@
...
+  become: yes
+  unarchive:
+    # src: "https://packages.cnx.org/js-builds/webview_{{ webview_version}}.tar.gz"
+    # remote_src: yes
+    src: "{{ temp_webview_tarball_location.stdout }}"
+    copy: no
+    # ^^^ changes needed to make this work until the ansible 2.2
+    dest: "{{ temp_webview_location.stdout }}"
+    owner: www-data
+    group: www-data
```
karenc: Is it not possible to use the url directly?  Like https://github.com/Connexions/cnx-deploy/blob/fabd873dfc07d33ef5be5f79124a08521646865b/tasks/install_python24.yml#L80

pumazi:
@karenc thoughts now?

PR #112, opened 2016-10-11
Add the production environment

pumazi:


pumazi:
This has now grown to 5 boxes:
 1. initial request load balancer
 1. front-end cache and webserver
 1. front-end cache and webserver
 1. zope front-end load balancer & cache
 1. nfs server

PR #123, opened 2016-10-20
Fix configure_hosts playbook to include new host groups

pumazi:
This is just a rather simple fix to include the new and existing host groups in the configure_hosts playbook. The nfs host group needed to be added to this playbook in order to mock what we are deploying, which is a dedicated NFS server.


pumazi:
Included in #112

PR #133, opened 2016-11-01
Fix a typo in the variable for the concept coach webview settings

pumazi:
I couldn’t decide what to call the variable, because honestly I had no idea what the setting was for. I’m still not altogether certain what the mapping is for. :stuck_out_tongue:

### DIFF
--- a/environments/prod-phase-one/group_vars/all/vars.yml
+++ b/environments/prod-phase-one/group_vars/all/vars.yml
@@ -55,7 +55,7 @@ xxx_hardcoded_zclients:
 
 blocked_ip_addresses: "{{ vault_blocked_ip_addresses }}"
 
-concept_coach_webview_mapping:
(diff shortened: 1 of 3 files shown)


dtwilliamson:
typo-fix needs a typo fix

environments/prod-phase-one/group_vars/all/vars.yml
```diff
@@ -55,7 +55,7 @@ xxx_hardcoded_zclients:
 
 blocked_ip_addresses: "{{ vault_blocked_ip_addresses }}"
 
-concept_coach_webview_mapping:
+concept_couch_webview_settings:
```
dtwilliamson: "couch"

pumazi:
Ok, fixed up. It's now using the correct variable name in all the right places. :)

PR #139, opened 2016-11-05
Zope components as roles

pumazi:
This creates roles for each of the components (`zeo`, `zclient` and `pdf_gen`). There is a common role created, called `zope_common`, which is used to get the source and build the application. These are all pull together via the meta role, called `zope`.


pumazi:
The easiest way to review this is probably to view it one commit at a time. Each commit makes a minor adjustment to the `main.yml` playbook. Each commit is intended to work, so nothing is broken between the commits, which means it should read like a horror story. ;)

pumazi:
I'll need to add the zope group of groups to the various environment inventory files.

```ini
[zope:children]
zeo
zclient
pdf_gen
```

pumazi:
Updated with the group of groups declaration in dev and vm environments. I didn't bother with cte-dev because it's stuck on a specific changeset and will likely be deleted in the future.

PR #166, opened 2016-12-15
Tutor staging

dtwilliamson:
change staging00 to staging and point to tutor-staging

### DIFF
--- a/environments/staging-phase-one/group_vars/all/vars.yml
+++ b/environments/staging-phase-one/group_vars/all/vars.yml
@@ -26,7 +26,7 @@ zope_domain: legacy-staging00.cnx.org
 arclishing_domain: archive-staging00.cnx.org
 frontend_domain: staging00.cnx.org
 accounts_domain: accounts.openstax.org
-tutor_domain: tutor.openstax.org
+tutor_domain: tutor-staging.openstax.org
 exercises_domain: exercises.openstax.org
 
 accounts_consumer_token: "{{ vault_accounts_consumer_token }}"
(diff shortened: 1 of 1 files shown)


environments/staging-phase-one/group_vars/all/vars.yml
```diff
@@ -22,11 +22,11 @@ publishing_port: 6550
...
 
-zope_domain: legacy-staging00.cnx.org
-arclishing_domain: archive-staging00.cnx.org
-frontend_domain: staging00.cnx.org
+zope_domain: legacy-staging.cnx.org
+arclishing_domain: archive-staging.cnx.org
+frontend_domain: staging.cnx.org
 accounts_domain: accounts.openstax.org
-tutor_domain: tutor.openstax.org
+tutor_domain: tutor-staging.openstax.org
```
pumazi: I’m not exactly in favor of this kind of arrangement. I think the cnx (and tutor) staging environment should point at the production services that are out of their scope. Otherwise, we end up tightly coupling changes to services that may or may not be deployed (deadlock scenario). That means when two or more services need deployed at the same time, which isn’t usually a good idea because there is a lot more risk involved, we will continually run into chicken and egg issues. In this case, cnx comes before tutor, but that may not always be the case.

pumazi: Let’s assume we will never do a tutor and cnx deployment at the same time, because that’s a silly amount of risk. Say you’ve got a change to cnx that we have out on staging, call it A1. Tutor is also working on changes, call those B1.  Now say A1 is prepped for release on staging.cnx.org and tutor-staging is pointing at that environment. For whatever reason a problem pops up in A1, which means the deploy of A1 is a no go. This now discounts any tests that might depend on cnx in the tutor environment, because the problem could be effecting that environment, which now means B1 is a no go. Had tutor’s staging environment been using production, the release and deploy might have gone through.

PR #176, opened 2017-01-05
Add a cron job to count content hits

### LINKED ISSUE #165: Create cronjob to run the hits counter on a weekly basis
## Summary

The hits counter script (`cnx-archive-hits_counter`) consumes the varnish log and counts the url hits for content. This information is then used to weight the search by popularity. The script needs to be run weekly to consume and update the data within the database.

## Environment

all

pumazi:
Closes #165

### DIFF
--- a/roles/archive/templates/etc/cron.weekly/count_archive_hits
+++ b/roles/archive/templates/etc/cron.weekly/count_archive_hits
@@ -1,20 +1,37 @@
-#!/bin/sh
-set -e
+#!/bin/bash
+#set -e
 
-INSTANCE_LOGS_ROOT=/var/www/varnish_logs
+log_root=/var/log/varnish
+log_user=varnishlog
+log_group=varnish
+combined_out="${log_root}/varnishncsa_combined.log"
+current_datestamp=$(date -d "$(printf '%(%m/%d/%Y)T 00:00:00')" +%s)
+timestamp_file="${log_root}/varnish_last_import"
+varnfiles_tmp=$(mktemp /tmp/varnish.XXXXXX)
 
+# Run find command to create a temp file to use for combining logs
+# into one file.
+# If you are wondering why +20c it's because an empty gz file is 20 bytes
+if [[ ! -e $timestamp_file ]] ; then
+    find ${log_root}/ -mindepth 1 \( \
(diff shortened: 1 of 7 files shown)


pumazi:
`/etc/cron.daily/copy_varnish_logs` will be installed an run on servers that contain a varnish instance (e.g. staging{01,02,03}).

`/etc/cron.weekly/count_archive_hits` will be installed on the database server, which contains a copy of the archive codebase for triggers. This doesn't need to be run on the database server, but it seems to me that would the the easiest way to ensure there is only one instance of it out there, unless we can prevent a race condition from happening (maybe creating the timestamp file before processing the data and have a decent cleanup procedure).

pumazi:
@DarkPrinceFrost The logic for command that gets call can be found at https://github.com/Connexions/cnx-archive/blob/master/cnxarchive/scripts/hits_counter.py Though it's probably not necessary to know the details, but just in case you wanted to know. 😃

pumazi:
@dtwilliamson This is all yours to review and merge. 😄

roles/archive/templates/etc/cron.weekly/count_archive_hits
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#set -e
+
+log_root={{ varnish_logroot|default('/var/www/varnish_logs') }}
+log_user={{ varnish_user|default('varnishlog') }}
+log_group={{ varnish_group|default('varnish') }}
+combined_out="${log_root}/varnishncsa_combined.log"
+current_datestamp=$(date -d "$(printf '%(%m/%d/%Y)T 00:00:00')" +%s)
```
dtwilliamson: date can do that by itself: `date -d '00:00:00' +%s` - you can add "today" if you want to make it more readable: `date -d 'today 00:00:00' +%s`.

roles/archive/templates/etc/cron.weekly/count_archive_hits
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#set -e
+
```
dtwilliamson: I recommend always quoting variables by habit even if the values aren't expected to include whitespace.

PR #179, opened 2017-01-10
Iptables

### LINKED ISSUE #4: Lock down the Zope app at the TCP level
The Zope server app needs to be blocked to all income requests except those from the frontend server, which is proxying the requests through varnish. In production this is done via setting up an iptables rule that only allows the frontend servers to establish an http connection.

DarkPrinceFrost:
Closes #4


roles/iptables/tasks/iptables.yml
```diff
@@ -0,0 +1,21 @@
...
+  when: open_tcp_ports is undefined
+
+- name: iptables chain setup
+  become: yes
+  script: ../scripts/iptables_setup.sh
+
+- name: iptables template setup
+  become: yes
+  template:
+    src: templates/iptables.j2
```
pumazi: Can you drop the `.j2` from the template filename.

roles/iptables/tasks/rsyslog.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: rsyslog conf template
+  template:
+    src: templates/rsyslog.conf.j2
```
pumazi: ... `.j2`

roles/iptables/tasks/rsyslog.yml
```diff
@@ -0,0 +1,15 @@
...
+    dest: /tmp/rsyslog.conf
+    mode: 0644
+  tags: iptables-rsyslog-conf
+
+- name: rsyslog iptables template
+  template:
+    src: templates/40-iptables.conf.j2
```
pumazi: ... `.j2`

roles/iptables/tasks/rsyslog.yml
```diff
@@ -0,0 +1,15 @@
...
+    dest: /tmp/40-iptables.conf.j2
```
pumazi: The destination probably shouldn't have the `.j2` in it. 😛

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -325,6 +325,9 @@ sub vcl_miss {
 }
 
 sub vcl_backend_response {
+    if (bereq.url ~ "^/scripts/(?:main|settings)[.]js$") {
```
pumazi: This change is already in. Maybe you need to rebase against origin/master?

roles/iptables/tasks/iptables.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: iptables variables
+  include_vars: vars/iptables.yml
```
pumazi: This task is probably unnecessary. Try placing the iptables.yml in the role as `roles/iptables/defaults/main.yml`.

roles/iptables/tasks/iptables.yml
pumazi: Also, can you put the template at the same filesystem location as the destination? For example, take a look at the archive role's template task.

roles/iptables/tasks/iptables.yml
dtwilliamson: I took care of this in the rsyslog branch (PR 182)

pumazi:
@DarkPrinceFrost @dtwilliamson : Can you have a look at this to ensure all the things are still correct? I've removed the merge commits, squashed a few things and touched up minor elements. I correct an typo with the new `zclient_base_port` variable usage as well as committed a change to use that variable elsewhere.

pumazi:
I'm looking forward to having the rsyslog foundation in place. 💯

pumazi:
I've been able to run the `iptables.yml` playbook locally. I haven't verified the results beyond being able to run it. I figure we can see the results on the dev server and you guys can ensure it works there. 😄

Issue #186, opened 2017-01-17
Load balance authoring


pumazi:
## Summary

We need to add a load balancer for the authoring instances. Both archive and publishing are load balanced by varnish, because their requests go through varnish. Authoring's requests go through varnish as well, but only as a proxy to nginx. So the likely place to load balance authoring is within nginx.

## Environment

beta

PR #187, opened 2017-01-18
Load balanced cnx-suite apps

pumazi:
Depends on #185 

💀  **Theses changes are not backwards compatible**. 💀  (The last commit breaks production.) These changes require the cnx-suite to be deployed and managed by cnx-deploy. Acceptance of this pull request means that we will be required to fork a stability version of cnx-deploy (done: see `production` branch) until [Phase Three](https://github.com/Connexions/cnx-deploy/milestone/3) hits production.

This is missing a load balancing element for authoring (see also #186) . But authoring is not used in production and will continue to function regardless of these changes. As such, I've chosen to push this work into the future.

### DIFF
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -1,6 +1,7 @@
 vcl 4.0;
 
 import std;
+import directors;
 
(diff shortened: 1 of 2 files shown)


pumazi:
@dtwilliamson @DarkPrinceFrost : This is ready for review and merge. 😸

PR #190, opened 2017-01-24
Added new qa server to inventory

DarkPrinceFrost:


roles/webview/tasks/google_site_verification.yml
```diff
@@ -19,3 +21,6 @@
     owner: www-data
     group: www-data
     mode: 0744
+  when: google_site_verification_stat.stat.exists
```
pumazi: This adds a second `when` to the task.

roles/webview/tasks/google_site_verification.yml
```diff
@@ -8,7 +8,9 @@
     path: "{{ inventory_dir }}/secrets/google/{{ google_site_verification_filename }}"
   register: google_site_verification_stat
   # Fail when the filename is defined but the file does not exist.
-  failed_when: "not google_site_verification_stat.stat.exists"
+  # failed_when: "not google_site_verification_stat.stat.exists"
```
pumazi: Take a look at the `failed_when` documentation.

pumazi:
The qa env that this is being revised is out dated. There are missing groups and likely missing settings. It's probably best to remove it and copy from either dev or staging.

PR #194, opened 2017-01-27
Pin versions based on what's in production at the moment

pumazi:
I believe these version pinnings reflect what is in production, but you may want to verify that.


environments/staging-phase-two/files/versions.cfg
```diff
@@ -0,0 +1,99 @@
...
+Products.RhaptosCatFixes = 0.2
+Products.RhaptosCollaborationTool = 0.10.1
+Products.RhaptosCollection = 0.75
+Products.RhaptosContent = 0.66
+Products.RhaptosHitCountTool = 0.12
+Products.RhaptosModuleEditor = 0.83
+Products.RhaptosModuleStorage = 1.1.2
+Products.RhaptosPDFLatexTool = 0.15.1dev
+Products.RhaptosPatchTool = 0.7.2
+Products.RhaptosPrint = 0.90
```
pumazi: There may have been a release of RhaptosPrint since I last updated this file. This is likely one version pinning to verify.

environments/staging-phase-two/group_vars/all/vars.yml
```diff
@@ -45,7 +45,7 @@ nfs_server_for_specials: staging-files00.cnx.org
 
 nfs_connections: "{{ vault_nfs_connections }}"
 
-webview_version: v0.21.0
```
DarkPrinceFrost: So we don't want to use `latest` here?

pumazi: No, but it should be updated to what's in production, which is currently `v0.21.11-dev7fefc3d`. How a development version hit production is beyond me.

pumazi: O... this is already correct. 👍 

We could have potentially used latest-stable (latest tagged version), but since we are using a development build in production, we can't do that. 😖

PR #221, opened 2017-02-09
Added redirect to remove www from host

DarkPrinceFrost:


pumazi:
As written, this will not redirect the user, but instead this allows the `www*.cnx.org` name to be used. AFAICT, this rewrites the host header. I just wanted to point that out. **Is everyone ok with this being a rewrite rather than a redirect?**

It's likely not an issue, but the user will likely be forcibly directed to the non-www after surfing through the site. This would be done for example by clicking one of the links within a page's content. This could lead to mysterious issues, but things should be ok.

pumazi:
Closing in favor of redirecting rather than rewriting. (this decision comes out of a discussion in slack)

PR #225, opened 2017-02-15
Enable haproxy to set a cookie to allow for instance affinity

pumazi:
This means client requests will be directed to the same application instance
instead of bouncing them around.


pumazi:
That's not exactly a change request. Instead it's more like saying don't do this. 😝 

I was just translating what you guys did while I was out. I'm cool with leaving this commented out.

reedstrm:
@pumazi - yup, github's review option doesn't have a "REJECTED!" button :)

PR #232, opened 2017-02-17
Postgres enable

DarkPrinceFrost:
Fixed typo and added a fix for https://github.com/Connexions/cnx-deploy/issues/75

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -9,6 +9,9 @@ default_archive_base_port: 6500
 default_publishing_base_port: 6600
 default_authoring_base_port: 6400
 
+# See http://docs.ansible.com/ansible/postgresql_user_module.html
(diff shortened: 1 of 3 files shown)


group_vars/all.yml
```diff
@@ -9,6 +9,9 @@ default_archive_base_port: 6500
 default_publishing_base_port: 6600
 default_authoring_base_port: 6400
 
+# See http://docs.ansible.com/ansible/postgresql_user_module.html
+# Change in group_vars or host_vars for environment specific settings
+role_attr_flags: CREATEDB,NOSUPERUSER
```
pumazi: I'm not exactly sure this is going to be enough since cnx's database creation logic likes to be a superuser, but we'll roll with it. If it fails later, we'll have at least tried. 👍

PR #235, opened 2017-02-20
change how we detect existing libreoffice4.1 install - survives syste…

reedstrm:
…m restart

### DIFF
--- a/roles/zclient/tasks/main.yml
+++ b/roles/zclient/tasks/main.yml
@@ -19,6 +19,7 @@
     - install-libreoffice
 
 - name: install libreoffice
+  when: "not libreoffice_installed_python.stat.exists"
   become: yes
   shell: "dpkg -i *.deb"
   args:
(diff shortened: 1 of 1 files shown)


roles/zclient/tasks/main.yml
```diff
@@ -2,15 +2,15 @@
...
-    path: "/tmp/LibreOffice_4.1.6.2_Linux_x86-64_deb"
-  register: libreoffice_download_location
+    path: "/opt/libreoffice4.1/program/python"
+  register: libreoffice_installed_python
   tags:
     - install-libreoffice
 
 - name: unpack libreoffice tarball
-  when: "not libreoffice_download_location.stat.exists"
+  when: "not libreoffice_installed_python.stat.exists"
```
pumazi: Can you also put this 'when' on the task immediately following this one?

pumazi: Sorry, it the 'when' wasn't needed on the followup task before, but would certainly be required now. Otherwise, great change. 👍

PR #237, opened 2017-02-21
Install unoconv

DarkPrinceFrost:


roles/zclient/templates/etc/supervisor/conf.d/zclients.conf
```diff
@@ -9,7 +9,7 @@ stopwaitsecs=60
 
 
 [program:unoconv]
-command=/var/lib/cnx/cnx-buildout/scripts/unoconv -l
+command=/usr/local/bin/unoconv -l
```
reedstrm: probably needs to be `python3 /usr/local/bin/unoconv -l`

reedstrm: seems we can use the system/distro unoconv after all, so just `unoconv -l`

reedstrm: unless it's best practice to use full paths, then `/usr/bin/unoconv -l`

PR #249, opened 2017-03-09
quoting values in app.ini breaks them

reedstrm:
Having quotes around these values (silently) breaks exercise fetching and TeX math conversion. Oops.


pumazi:
We'll need a PR for cnx-publishing as well: https://github.com/Connexions/cnx-publishing/blob/2e407150be4e3207546dbd76a6900cf64076feb0/development.ini#L26-L29

PR #253, opened 2017-03-10
Define a global default zclient_base_port

pumazi:


dtwilliamson:
default_zclient_base_port and base_port names correspond?

PR #265, opened 2017-04-05
reduce HSTS level to what we can handle for now

reedstrm:
This brings cnx.org in line with what we've serving for openstax.org, and remove the requirement that all subdomains (like dev.cnx.org or other test and development machines) be served exclusively over https


dtwilliamson:
Once this is merged, how do we want to handle deployment to production?

PR #270, opened 2017-04-07
Remove the postgres install for zope machines

pumazi:
We previously installed postgres on the machine used by zclients so that
the site creation could complete. We did this because the site creation
needs an empty database in order to successfully run. Later we transitioned
to creating a dummy database on the database server for it to use.
However, we forgot to remove the postgres install from the previous solution.
This takes care of removing that previous solution.

### DIFF
--- a/database.yml
+++ b/database.yml
@@ -10,18 +10,6 @@
 
 - name: install postgres database
   hosts:
-    # FIXME The zclient group's rhaptos site creation requires a local database.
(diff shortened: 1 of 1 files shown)


pumazi:
@dtwilliamson this should remove the necessity to create the database role again during rhaptos_site_creation, because it forces the script to use the remote database rather than the locally running instance.

PR #271, opened 2017-04-07
Remove authoring location from webview nginx site conf when unavailable

pumazi:
This removes the authoring locations from the webview nginx configuration
when authoring is not defined or used, which is the case for production.

This does clear up an issue with 502's on certain routes on production.


pumazi:
@dtwilliamson this should remove the need for the `authoring_host` variable. This should resolve the last issue of needing to define the `<app>_host` variables.

PR #283, opened 2017-04-14
Define host prefix for the qa environment

pumazi:
Submitting @DarkPrinceFrost's qa environment modifications. I changed the `host_prefix` var to a (by convention) private variable, `_host_prefix`. This is now independent of the iptables changes.


dtwilliamson:
will we use this prefix in all environments? It strikes me as greatly simplifying the creation of new environments.

pumazi:
@dtwilliamson unfortunately, no. This is simplified here simply because qa is a single host deployment. But the prefix is not always the same across the board for the settings this specific change touches.

dtwilliamson:
Ok, I see. What about a two-part prefix? Along the lines of

p1=ENV; p2_auth=00; p2_arch=01; authoring_db_host: "{{ p1 }}{{ p2 }}.cnx.org"

PR #294, opened 2017-04-19
add archive restart workaround script to cron

dtwilliamson:


dtwilliamson:
Changes made to support new-style cnx-deploy installation

PR #302, opened 2017-04-21
add vars.sh files to hold shell vars for script use

dtwilliamson:
I changed the way the with_items works to reduce repetition and changed the file perms from 744 to 644 since I didn't think they need to be executable.

The use of the variables in the archive-restart script will come in a separate PR.

### DIFF
--- a/roles/archive/tasks/main.yml
+++ b/roles/archive/tasks/main.yml
@@ -112,14 +112,14 @@
 - name: render configuration
   become: yes
   template:
-    src: "{{ item.src }}"
-    dest: "{{ item.dest }}"
-    mode: 0744
+    src: "{{ item }}"
+    dest: "/{{ item }}"
+    mode: 0644
     owner: root
     group: root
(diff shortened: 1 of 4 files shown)


roles/archive/tasks/main.yml
```diff
@@ -112,14 +112,14 @@
...
+    src: "{{ item }}"
+    dest: "/{{ item }}"
+    mode: 0644
     owner: root
     group: root
   with_items:
-    - {src: "etc/cnx/archive/app.ini",
-       dest: "/etc/cnx/archive/app.ini"}
+    - "etc/cnx/archive/app.ini"
+    - "etc/cnx/archive/vars.sh"
```
pumazi: What happened to app.ini? Perhaps you meant to copy this task before you made this change?

pumazi: Ah Ok, I see it now. 👍

PR #310, opened 2017-05-12
Fixes that were necessary for a successful deployment

pumazi:
These are changes we had to make in order for the Phase Three deploy go through.

### DIFF
--- a/roles/postgres_replicant/tasks/main.yml
+++ b/roles/postgres_replicant/tasks/main.yml
@@ -39,6 +39,7 @@
       command: "pg_basebackup --write-recovery-conf --pgdata=/var/lib/postgresql/{{ postgres_version }}/main --host={{ hostvars[groups.database[0]].ansible_default_ipv4.address }} --port=5432 --username={{ replicator_db_user|default(default_replicator_db_user) }} --xlog-method=stream --checkpoint=fast"
 
 - name: configure recovery.conf
+  become: yes
   template:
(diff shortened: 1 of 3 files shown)


roles/postgres_replicant/tasks/main.yml
```diff
@@ -47,6 +48,7 @@
     mode: 0640
 
 - name: start postgresql service
+  become: yes
   service:
     name: postgresql
```
dtwilliamson: The change I made when we did the deploy was to change the service name to postgresql@9.4-main.service

swap_database.yml
```diff
@@ -82,10 +82,12 @@
...
     - name: copy database from master to replicant
+      become: yes
       become_user: postgres
       command: "pg_basebackup --write-recovery-conf --pgdata=/var/lib/postgresql/{{ postgres_version }}/main --host={{ hostvars[groups.database[0]].ansible_default_ipv4.address }} --port=5432 --username={{ replicator_db_user|default(default_replicator_db_user) }} --xlog-method=stream --checkpoint=fast"
     # Touch recovery.conf to prevent data removal during postgres_replicant's role execution.
     - name: touch recovery.conf
+      become: yes
       file:
         path: "/var/lib/postgresql/{{ postgres_version }}/main/recovery.conf"
         owner: postgres
```
dtwilliamson: Line 94 has a typo:

        grouap: postgres
 should be
        group: postgres

PR #317, opened 2017-05-31
Variable and group fixes for dev, tea and qa envs

pumazi:


environments/qa/inventory
```diff
@@ -11,6 +11,9 @@ qa00
 [database:children]
 qa00
 
+[replicant:children]
+# None, because single-machine install
```
reedstrm: Oh, is this sufficient to have it not do the replication at all?

pumazi: From cnx-deploy chat:
> mulich: Ha! We've apparently set things up correct the first time around. Leaving the replicant group empty seems to just work.

PR #319, opened 2017-05-31
Varnish fixme

DarkPrinceFrost:


roles/varnish/templates/etc/default/varnish
```diff
@@ -0,0 +1,10 @@
+START=yes
+NFILES={{ varnish.nfiles|default(131072) }}
+MEMLOCK={{ varnish.memlock|default(82000) }}
+DAEMON_OPTS="-j unix,user=vcache -F \
+             -a :{{ varnish.port }} \
+             -T 127.0.0.1:6082 \
+             -f /etc/varnish/varnish.vcl \
+             -S /etc/varnish/secret \
+             -s malloc,{{ varnish.malloc|default('16G') }} \
```
pumazi: Perhaps we should make these default `256m`, which is the default on install. It seems the reversal of what we have been doing, which was leave it low and then set it high when needed.

Looks like you are applying this default twice too.

roles/varnish/vars/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+varnish:
+  user: varnishlog
+  group: varnish
+  port: "{{ varnish_port|default(80) }}"
+  malloc: "{{ varnish_malloc|default('16g') }}"
```
pumazi: Default `256m`?

roles/varnish/vars/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+varnish:
+  user: varnishlog
+  group: varnish
+  port: "{{ varnish_port|default(80) }}"
+  malloc: "{{ varnish_malloc|default('256m') }}"
```
pumazi: The current variable that is out there in environments is `varnish_memory_allocation`.

environments/prod-phase-three/group_vars/all/vars.yml
```diff
@@ -10,6 +10,7 @@ archive_db_port: 5432
 memcached_hosts: localhost
 
 varnish_port: 80
+varnish_malloc: 16g
```
pumazi: This is set at the `host_vars` level (see `host_vars/prod0{1,2,3}`).

PR #320, opened 2017-05-31
insure we can make a replication connection before blowing away local db

reedstrm:
This is copy-pasta from the similar check to make sure we're not already running a replicant. I checked the command locally but haven't tried the playbook

### DIFF
--- a/roles/postgres_replicant/tasks/main.yml
+++ b/roles/postgres_replicant/tasks/main.yml
@@ -48,7 +48,11 @@
       become_user: postgres
       command: "pg_basebackup --write-recovery-conf --pgdata=/var/lib/postgresql/{{ postgres_version }}/main --host={{ hostvars[groups.database[0]].ansible_default_ipv4.address }} --port=5432 --username={{ replicator_db_user|default(default_replicator_db_user) }} --xlog-method=stream --checkpoint=fast"
 
-- name: configure recovery.conf
+- when:
+    - "'f' in replication_running.stdout"
+    - not recoveryconf_file.stat.exists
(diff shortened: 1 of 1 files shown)


roles/postgres_replicant/tasks/main.yml
```diff
@@ -18,9 +18,19 @@
...
+  command: psql 'dbname=postgres replication=database host={{ hostvars[groups.database[0]].ansible_default_ipv4.address }} user={{ replicator_db_user|default(default_replicator_db_user) }}' -c 'IDENTIFY_SYSTEM;'
+  register: replication_connection
+  tags:
+    - database
+    - replication
+
 - when:
     - "'f' in replication_running.stdout"
     - not recoveryconf_file.stat.exists
+    - "'systemid' in replication_connection.stdout"
```
pumazi: You'll need to add this to the next task as well, no? Because if this doesn't run and the following task does, the `recovery.conf` will be created. This means that on next run, this would fail the condition for another reason, which is probably not the intended case.

reedstrm: Ah, yes, would really like a 'bail out if this fails' Is there a way to spell that?

pumazi: Hmm... is it possible that the main service hasn't fully restarted yet and that is why it's not ready for the replication connection?

You can deliberately stop the play using the [fail module](https://docs.ansible.com/ansible/fail_module.html), but we don't really want to do this because it can interrupt handlers that may have already been triggered. Though, that is not the case for this role at this time.

dtwilliamson:
Are they not? It looks like they are.

PR #321, opened 2017-05-31
Fix to use a local connection rather then the local host definition

pumazi:
The `local` host definition was removed several changesets ago. This one must have slipped through. This removes it in favor of using the localhost host and a local connection.


pumazi:
See also, https://docs.ansible.com/ansible/playbooks_delegation.html#local-playbooks

Issue #274, opened 2017-04-12
webview version mismatch


pumazi:
## Summary

We've run into a issue with the version.txt within webview. The problem arose because the intuitive thing to do from the DevOps perspective didn't match up with what actually happens. So we need to fix that.
It was believed that the webview tarball didn't need to be recreated since the tag from `v0.22.1rc4` is equal to `v0.22.1`. Thus, why rebuild it? Well, the problem is that the tarballs may be build from the same changeset but they come out with one significant difference, the contents of `version.txt`.

You can see an example of this:
```
$ http -b http://cnx.org/version.txt
v0.22.1rc4-dev74f5709
```

We are going to fix this minor difference by simply removing it. We will now be rendering the `version.txt` file during deployment instead of at build/compilation time.

### TODO

- [x] remove `version.txt` from the tarball build script (@philschatz)
- [ ] remove the `latest` and `latest-stable` symlinks from {builds,packages}.cnx.org (@DarkPrinceFrost | @dtwilliamson | @pumazi)
- [x] put a sha bit in the webview code, if possible; otherwise put it in a `rev.txt` (@philschatz)
- [x] remove the `latest` and `latest-stable` symlink creation from the build script (@philschatz)
- [x] remove tasks that use `latest*` in cnx-deploy (@pumazi | @DarkPrinceFrost | @dtwilliamson)
- [x] create tasks to render the `version.txt` in cnx-deploy (@pumazi | @DarkPrinceFrost | @dtwilliamson)

## Environment

all

philschatz:
I had trouble using `git archive` to include the contents of the `dist/` directory it seems like it does not include files that are in `.gitignore`.

The commands I used were (from https://git-scm.com/docs/git-archive ):

```sh
mkdir ../deployment-testing
echo 'Last commit date: $Format:%cd by %aN$' > LAST_COMMIT
touch .gitattributes
git add LAST_COMMIT .gitattributes
git commit -am 'adding LAST_COMMIT file for archives'

mkdir ../deployment-testing
git archive HEAD dist/ | tar xCf ../deployment-testing -
```

Any ideas on what I'm doing wrong?

pumazi:
[...]

PR #325, opened 2017-06-01
Use a webview tarball without an included version.txt

pumazi:
This works towards fulfilling the cnx-deploy obligations of #274.

### DIFF
--- a/environments/dev/group_vars/all/vars.yml
+++ b/environments/dev/group_vars/all/vars.yml
@@ -50,7 +50,7 @@ nfs_server_for_files2: dev00.cnx.org
 nfs_server_for_specials: dev00.cnx.org
 nfs_server_for_varnish_logs: dev00.cnx.org
 
-webview_version: latest
+webview_version: master
 
 concept_coach_webview_settings:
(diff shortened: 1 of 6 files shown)


philschatz:
Per some discussion from way back on April 12, the `webview_*.tar.gz` will begin having a `/rev.txt` instead of a `/version.txt` in it.

Jenkins has been running 2 builds in order to test that feature:

- "Build CNX Webview"
- "temp_Build CNX Webview"

The `temp_` one does the following things differently:

1. generates the new `/rev.txt` file
1. build a `_temp_webview_*.tar.gz` file
1. update the `webview_master.tar.gz` symlink when the master branch is built

When folks think it's ok, I can do the following:

1. remove the old Jenkins build
1. have the new build generate `webview_*.tar.gz` (without the `_temp` prefix)

/cc @reedstrm

philschatz:
@pumazi I thought `version.txt` was going to contain the versions of all the repos that are deployed, not just webview. Right?

pumazi:
@philschatz Nope. We are just moving what creates the `version.txt`. We may eventually expose the requirements.txt files. But that's different from this.

PR #326, opened 2017-06-01
start postgres instead of calling the replicant role

dtwilliamson:


swap_database.yml
```diff
@@ -96,10 +96,12 @@
         state: touch
     # The replicant service should still be stopped at this point.
```
reedstrm: Hmm, for non-single-server environments, doesn't the postgresql service need to be restarted on replicant?

dtwilliamson: I forgot to add a `hosts:` clause. Thanks.

dtwilliamson:
Actually, I should probably just remove the touch step, since it's not needed at all.

pumazi:
So, replication setup for the swapped in database is now a manual operation?

dtwilliamson:
No, swa_database.yml includes a pg_basebackup step and everything else necessary. If there's a better approach to making all this work, I'd like to know. I'm in the process of testing the changes, btw.

pumazi:
I thought the replicant would end up unconfigured, which is why the role is called. Doesn't pg_basebackup blow away the existing data directory?

dtwilliamson:
The first step of the role requires that postgres is running in order to respond to the query. Besides, the replicant role has already been run when the replicant server was provisioned. At that time, though, the primary database is empty since it's newly provisioned. copy_database.yml copies the db from prod to staging, for example, as dbname_new on the primary and does nothing with the replicant. swap_database.yml, swaps dbname to _old and _new to dbname on the primary then does pg_basebackup from the primary to the secondary. So I wouldn't say the replicant is unconfigured since it was previously provisioned and then becomes a duplicate of the primary.

PR #329, opened 2017-06-06
variable added and slack ignore errors

DarkPrinceFrost:


dtwilliamson:
The ignore directive needs to be available generally. I don't know why it was put in this tea branch.

PR #339, opened 2017-06-29
Add the publishing worker process

pumazi:
This adds two new host groups: `publishing_worker` and `broker`. (It also removes the more generic and unused `worker` host group.)

One of the added components is a message queue, RabbitMQ. The other is a publishing worker process for churning on asynchronous workflow tasks.

Note, I have not included updates to every environment in this set of changes. So that will need to happen as a followup. You can see the gist of what that will look like within the `environments/vm` changes here.


pumazi:
This doesn't yet work for multi-box setups, but should work just fine for single-box deploys. I have a feeling we'll need to look deeper into the configuration of rabbitmq to solve that. But let's not let that hold us up. In the mean time, I'll work towards resolving the issue.

pumazi:
This is now running on `des`, which is a multi-system deploy. Connections between the procs are functional now.

PR #343, opened 2017-07-13
Make CNXML DTD pkg installs a dependent role

pumazi:
Instead of having this as it's own task make it a dependency of the roles that depend on this info.

This allows for more precision deployments via the role-based playbooks (e.g. `database.yml`).

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -23,18 +23,6 @@
 - include: database.yml
 - include: broker.yml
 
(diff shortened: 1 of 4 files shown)


karenc:
I tested this pull request by creating 3 VMs, one for cnx, one for database, one for zope.  This change correctly installs dtd packages on only database and zope.  So :+1:

PR #347, opened 2017-07-27
Update VM environment so it can set up a development environment

karenc:
I added a new variable called `venvs_owner`.  It is the owner of the `/var/cnx/venvs` directories.  I added it so it is possible to set the ssh user as the owner instead of www-data, but the default is still www-data if the variable is not set.  Having the ssh user as the owner in the VM environment makes it a lot easier for development if I need to change the code.

The other change is the one pumazi commented on: changing `vault_archive_db_name` and `vault_archive_db_user` to `archive_db_name` and `archive_db_user`.  The `vault_` variables should only be used in `group_vars/`, not directly in any templates or playbooks.

### DIFF
--- a/roles/archive/templates/etc/cnx/archive/vars.sh
+++ b/roles/archive/templates/etc/cnx/archive/vars.sh
@@ -1,8 +1,8 @@
 # this is a shell script which can be sourced into other scripts
 # to provide access to configuration values specific to this host
 
-archive_db_name="{{ vault_archive_db_name }}"
-archive_db_user="{{ vault_archive_db_user }}"
+archive_db_name="{{ archive_db_name }}"
+archive_db_user="{{ archive_db_user }}"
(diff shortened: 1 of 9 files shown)


roles/archive/templates/etc/cnx/archive/vars.sh
```diff
@@ -1,8 +1,8 @@
 # this is a shell script which can be sourced into other scripts
 # to provide access to configuration values specific to this host
 
-archive_db_name="{{ vault_archive_db_name }}"
-archive_db_user="{{ vault_archive_db_user }}"
+archive_db_name="{{ archive_db_name }}"
+archive_db_user="{{ archive_db_user }}"
```
pumazi: 🎈 🎉

pumazi:
@karenc : Create a small description for this so Dennis and Ryan know what's in it. Mostly because this doesn't just update the `vm` environment.

PR #351, opened 2017-07-31
Restart varnish to ensure it uses the correct port

### LINKED ISSUE #350: Varnish not running on the correct port
## Summary

When I first finished building the VM, /etc/default/varnish is correct but varnish is still using the default ports 6081 and 6082. Varnish is only restarted with the right ports when I change the `varnish_port` setting and run the playbook.

This means when I go to local.cnx.org, it shows a 503 Service Unavailable Error.

## Environment

VM on Ubuntu 16.04 (but probably affects all environments)

karenc:
Without this step, varnish was using the default port (6081) instead of
the one in /etc/default/varnish.

Close #350

### DIFF
--- a/roles/varnish/tasks/main.yml
+++ b/roles/varnish/tasks/main.yml
@@ -51,3 +51,11 @@
     src: "etc/cron.daily/copy_varnish_logs"
     dest: "/etc/cron.daily/copy_varnish_logs"
     mode: 0755
+
+# +++
+# Always Restart Varnish (to ensure Varnish is using the correct port)
+# +++
+- name: always restart varnish
+  command: /bin/true
+  notify:
+    - restart varnish


reedstrm:
Hmm, that file `/etc/default/varnish` mentions that it will be ignored if systemd  is running. Which it is on our servers.

DarkPrinceFrost:
Unless you use `EnvironmentFile=/etc/default/varnish` in `/lib/systemd/system/varnish.service`

karenc:
Do we need to restart varnish to clear the cache anyway?

karenc:
It looks like we needed to restart daemon tools to allow our service file changes available for use: 

https://github.com/Connexions/cnx-deploy/commit/6a4995ff3bb8aed01b1dc9bf993b4b7e6d5b20f4#diff-51527ad5cdd9e0878d2fb9127d53cb25R5

Closing this and submitting a different patch.

PR #354, opened 2017-08-04
Fix duplicate minor versions on QA and TEA

karenc:
When a page is published while the book is baking, we end up with two
copies of the book with exactly the same minor version.  While the code
for this is being fixed on cnx-archive, the data still needs to be
fixed.  This also includes temporarily disabling a trigger which fixes
the multiple minor versions problem, by updating the latest modules
table even when the book has not finished baking.

### DIFF
--- a/hotfixes/20170804_remove_duplicate_minor_versions.yml
+++ b/hotfixes/20170804_remove_duplicate_minor_versions.yml
@@ -1,0 +1,20 @@
+---
+# Used to fix multiple copies of the same minor versions on QA and TEA when
+# more than one page was published while the book was baking.
+
+- name: fix multiple copies of the same minor versions on QA and TEA
+  hosts: publishing
+  tasks:
+    - name: download python script to remove duplicate minor versions
+      get_url:
+        url: "https://raw.githubusercontent.com/Connexions/devops/master/scripts/remove_duplicate_minor_version_books.py"
+        dest: "/tmp"
+
+    - name: run python script to remove duplicate minor versions
+      shell: "/var/cnx/venvs/publishing/bin/python /tmp/remove_duplicate_minor_version_books.py -d 'dbname={{ archive_db_name }} user={{ archive_db_user }} password={{ archive_db_password }} host={{ archive_db_host }} port={{ archive_db_port }}'"
+
+- name: disable update_default_modules_stateid trigger
+  hosts: database
+  tasks:
+    - name: disable update_default_modules_stateid trigger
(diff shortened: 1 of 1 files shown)


reedstrm:
The intent is that we'll update the stateid later, manually, to trigger baking? And re-enable the trigger after we've fixed the code?

karenc:
From what I saw on QA, the result of disabling the `stateid` trigger is only that `latest_modules` is updated, I *think* `ruleset.css` in the book is still triggering baking so no manual baking was necessary.

hotfixes/20170804_remove_duplicate_minor_versions.yml
```diff
@@ -0,0 +1,17 @@
...
+  tasks:
+    - name: download python script to remove duplicate minor versions
+      get_url:
+        url: "https://raw.githubusercontent.com/Connexions/devops/master/scripts/remove_duplicate_minor_version_books.py"
+        dest: "/tmp"
+
+    - name: run python script to remove duplicate minor versions
+      shell: "/var/cnx/venvs/publishing/bin/python /tmp/remove_duplicate_minor_version_books.py -d 'dbname={{ archive_db_name }} user={{ archive_db_user }} password={{ archive_db_password }} host={{ archive_db_host }} port={{ archive_db_port }}'"
+
+    - name: disable update_default_modules_stateid trigger
```
karenc: Now that I think about it, we should probably run this on `database` instead of `publishing`.  In case `psql` is not installed on `publishing`...

reedstrm:
inserting a new ruleset.css triggers setting rebake(), which does sets stateid to 5 on modules.

karenc:
I don't mean inserting a new ruleset, I think the existing ruleset in the book seems to trigger rebake.

reedstrm:
Nope - only the insert - but a new minor version will get a new (link to) ruleset.css, so yes, that will trigger baking even w/o the default trigger. That trigger is for the future print_style based recipe selection.

karenc:
Yes, the disabled trigger is only for preventing an unbaked book from being in the `latest_modules` table.

karenc:
[[CTE.TEA1.10] Provide HTML Output of TEA Books to TEG/Gateway (CATCHALL Bug Fixes)](https://trello.com/c/VCgM57Ii/189-ctetea110-provide-html-output-of-tea-books-to-teg-gateway-bug-fixes)

karenc:
Doing the usual

```
ansible-playbook -i environments/qa/inventory hotfixes/20170804_remove_duplicate_minor_versions.yml
```

should work.

PR #358, opened 2017-08-14
Use the same version of webview on staging as prod

karenc:
This is already deployed to staging as [v2.1.4](https://github.com/Connexions/cnx-deploy/commits/v2.1.4)


karenc:
Will group this commit together with other commits in another PR.

PR #357, opened 2017-08-08
Create pdf_gen_common role for zclient and pdf_gen

karenc:
On the legacy site, PDFs are generated if they are not on the filesystem
when a user clicks on the download link, so zclients need to be able to
generate PDFs.

Create the pdf_gen_common role that both zclient and pdf_gen depend on
so that they both get all the packages and fonts necessary.

Close #353 

---

This has already been deployed to staging as [v2.1.3](https://github.com/Connexions/cnx-deploy/commits/v2.1.3).  Just need to add this to master.

PR #355, opened 2017-08-04
Add new NFS server for prod and staging

karenc:
We deployed to staging using [v2.1.2](https://github.com/Connexions/cnx-deploy/commits/v2.1.2), this is a pull request for master so we don't lose the changes.

PR #363, opened 2017-08-15
Changes in v2.1.4 - v2.1.10

karenc:
See https://github.com/Connexions/cnx-deploy/compare/v2.1.4...v2.1.10

v2.1.0 - v2.1.3 are in #355 and #357.

v2.1.9 are varnish changes from master. (so no need to include here)
v2.1.10 has been merged to master. (#364)

### DIFF
--- a/roles/publishing_common/tasks/main.yml
+++ b/roles/publishing_common/tasks/main.yml
@@ -110,7 +110,7 @@
   register: publishing_migrations_dir
 
 - name: migrate repository database
-  when: publishing_migrations_dir.stat.exists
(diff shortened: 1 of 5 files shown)


roles/publishing_common/tasks/main.yml
```diff
@@ -100,6 +100,20 @@
   pip:
     requirements: "/var/lib/cnx/publishing-requirements.txt"
     virtualenv: "/var/cnx/venvs/publishing"
+
+# FIXME this won't be necessary once publishing is >=0.7
```
pumazi: Should use BBB here, but that's fine.

PR #364, opened 2017-08-22
Use pyramid_sawing for archive logs

karenc:
This logs all requests and exceptions to archive in stdout with
timestamps.

For example:

```
192.168.122.250 - - [22/Aug/2017:12:19:29] "GET http://192.168.122.250/ HTTP/1.1" 200 OK 0 "-" "-"
2017-08-22 12:19:31,436 ERROR [waitress][waitress] Exception when serving /feeds/recent.rss
Traceback (most recent call last):
  File "/var/cnx/venvs/archive/local/lib/python2.7/site-packages/waitress/channel.py", line 338, in service
```

### DIFF
--- a/environments/prod/files/archive-requirements.txt
+++ b/environments/prod/files/archive-requirements.txt
@@ -17,6 +17,7 @@ psycopg2==2.5.4
 pyramid==1.6a2
 pyramid-jinja2==2.6.2
 pyramid-multiauth==0.8.0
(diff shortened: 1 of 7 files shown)


karenc:
@reedstrm this gives us logs with timestamps :)

karenc:
Yes, also we don't have any archive logs (except exceptions in stderr) at the moment... :p

PR #366, opened 2017-08-24
Redirect specials static file requests to https

karenc:
We are not sure about redirecting all traffic to https but it is
probably ok just for the specials static files to use https.

----

Tutor is somehow using the http links for the special files and the browser complains about having http resources... This sort of gets around the problem.  But we might want to find out why Tutor is using http links?


reedstrm:
Ed mentioned we've been notified from Google that we need to get everything under https by October anyway, so this might be superseded by a site-wide redirect soonish. Also agree that tutor shouldn't be using http - is it a config setting somewhere? @Dantemss @philschatz

karenc:
I don't think we're going to merge this.  I just created this in case tutor needed it last night.

PR #368, opened 2017-08-25
Fix static_files nginx config "try_files"

karenc:
nginx for static_files is returning "404 Not Found" for files in
/var/www/files2.

Move the regular expression to "location" instead.


karenc:
This is for v2.1.11: https://github.com/Connexions/cnx-deploy/compare/v2.1.10...v2.1.10-fix-static-files

PR #371, opened 2017-09-15
Move db-migrator to top of requirements.txt

karenc:
Pip doesn't have true dependency resolution (see
https://github.com/pypa/pip/issues/988).  For sub-dependencies that are
not specified in requirements.txt , pip uses the the definition of the
first package that specifies the dependency and considers the dependency
satisfied even if there are version conflicts.

In our case, on QA, the installed `psycopg2` is `2.6.2` and
`cnx-archive` requires `psycopg2>=2.5` which is ok, but
`db-migrator-1.0.1` requires `psycopg2>=2.7` which is ignored.

This change moves `db-migrator` to the top of requirements.txt so that
`psycopg2` is updated.

### DIFF
--- a/environments/__dev_envs/files/archive-requirements.txt
+++ b/environments/__dev_envs/files/archive-requirements.txt
@@ -1,10 +1,10 @@
+db-migrator==1.0.1
+
 -e git+https://github.com/Connexions/cnx-query-grammar.git#egg=cnx-query-grammar
 -e git+https://github.com/Connexions/rhaptos.cnxmlutils.git#egg=rhaptos.cnxmlutils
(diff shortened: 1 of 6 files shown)


pumazi:
Order shouldn't be a factor in the requirements.txt file. I'm fairly sure the requirements.txt file spec doesn't work on the premise of order listed in the file.

karenc:
I tried it locally by doing this:

```
pip install psycopg2==2.6.2
pip install -r archive-requirements.txt  # the old version
```

db-migrator is installed at 1.0.1, psycopg2 is still at 2.6.2 even though db-migrator requires >=2.7.  So it seems like the ticket from 2013 about pip not having true dependency resolution is still not fixed.

PR #374, opened 2017-09-20
Update prod and staging vars and inventory

karenc:
- Sync staging and prod envs from v2.1.13

  Copy changes for files/versions.cfg from v2.1.13.  All the other
  environment files are up to date.

- Update prod and staging inventory on master

  Update groups on prod and staging inventories by comparing them
  with the tea inventory.

  - Add `broker`, `channel_processing` and `publishing_worker` groups.
  - Remove `worker` group.
  - Add `channel_processing` and `publishing_worker` to `db_connected`
    group.

- Update ModuleStorage, Repository, ModuleEditor versions

  Update prod and staging to use the following versions:
   - Products.RhaptosModuleStorage 1.1.3
   - Products.RhaptosRepository 1.4.2
   - Products.RhaptosModuleEditor 0.85

PR #375, opened 2017-09-20
Pin versions for next release (of assignable webview)

pumazi:


pumazi:
Did I forget anything? cnx-db, cnx-epub, rhaptos.cnxmlutils, cnx-easybake, cnx-archive, cnx-publishing

The only other thing would be to check what versions of dependencies are installed on qa and dev that differ from those already in the requirements files.

karenc:
[[CTE.pipeline1.16] Deploy Assignable Webview to Production](https://trello.com/c/0BVWfU3T/175-ctepipeline116-deploy-assignable-webview-to-production)

reedstrm:
Also, the legacy side of this - should that be a separate PR?

karenc:
Most legacy changes are probably already deployed to prod but I'll check.  See also #374

karenc:
Production has more packages installed, but the versions of installed packages are the same between QA and Production.  So no need to do anything.

reedstrm:
Right, except for the two new PRs I asked for review of this morning, to fix the colliding collection ids …

PR #382, opened 2017-09-27
Redirect all cnx traffic to SSL

karenc:
Starting October 2017, Chrome (version 62) will show a "NOT SECURE"
warning when users enter text in a form on an HTTP page and for all HTTP
pages in incognito mode.

Our current PDF generation pipeline does not seem to work with HTTPS
urls (our xslt code is unable to fetch HTTPS url).  So haproxy will not
redirect traffic coming from the zope servers to use SSL.  Everything
else will be redirected.

### DIFF
--- a/lead_frontend.yml
+++ b/lead_frontend.yml
@@ -5,6 +5,7 @@
   hosts:
     - frontend
     - legacy_frontend
+    - zope
   # NOOP only to acquire IP address information about hosts.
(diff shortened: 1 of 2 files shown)


karenc:
[Make cnx.org use only HTTPS](https://trello.com/c/Ku3IEe27/240-make-cnxorg-use-only-https)

PR #383, opened 2017-09-27
Set rhaptos_print host to frontend domain

karenc:
The rhaptos_print tool in plone uses the `_host` attribute for building
zips and pdfs if it exists.  We need to set this to
`http://{{ frontend_domain }}` (not https) so xslt can fetch any urls it
needs when generating the files.


karenc:
[Make cnx.org use only HTTPS](https://trello.com/c/Ku3IEe27/240-make-cnxorg-use-only-https)

PR #384, opened 2017-09-28
Restart cnx apps when code is updated

karenc:
- Change celery supervisor group to publishing_worker

  The publishing celery workers supervisor group was "publishing", which
  is shared with the publishing pyramid application in another supervisor
  config file.  This does not seem to work when we do
  `supervisor restart publishing:`.  It restarts the publishing celery
  workers but not the publishing pyramid application.
  
  So we are giving the publishing celery workers a new group so we can
  continue to restart using groups.

- Restart cnx apps when code is updated

  Right now it seems our applications are only restarted when there is
  configuration change.  We need to restart the apps when our code change
  as well.
  
  For archive and authoring, it is straightforward, we just need to detect
  a config change or code change to restart the apps.
  
  For publishing, channel processing and publishing workers, we have to
  restart these processes the first time publishing is installed on the
  servers in order to detect whether code has been updated.  Since
  publishing_common is a prerequisite of all of the publishing roles, we
  don't know in advance which one is installed on the server.  So we are
  going to try to restart all of them if they are configured with
  supervisor.

### DIFF
--- a/roles/archive/tasks/main.yml
+++ b/roles/archive/tasks/main.yml
@@ -96,6 +96,8 @@
   pip:
     requirements: "/var/lib/cnx/archive-requirements.txt"
     virtualenv: "/var/cnx/venvs/archive"
+  notify:
+    - restart archive
 
 # +++
 # Configure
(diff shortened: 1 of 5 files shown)


karenc:
Yeah using handlers is so much nicer!  Thanks for the tip :smile:

pumazi:
Looks much cleaner! I guess the other advantage of handlers is that the changes become way easier to read. 😺

PR #389, opened 2017-10-03
Comment out the self signed cert in lead_frontend.yml

karenc:
Localhost `inventory_dir` is not defined causing this error when running
playbook `lead_frontend.yml`:

```
TASK [file] ******************************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"failed": true, "msg": "The task includes an option with an undefined variable. The error was: {{ inventory_dir }}/secrets/certs: 'inventory_d
ir' is undefined\n\nThe error appears to have been in '/home/karen/src/cnx-deploy/lead_frontend.yml': line 20, column 7, but may\nbe elsewhere in the file depending on the ex
act syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n    - file:\n      ^ here\n\nexception type: <class 'ansible.errors.AnsibleUndefinedVariable'>\nexception: {{ inventory_dir }}/secrets/certs: 'inventory_dir' is undefined"}
```

See bug report for ansible v2.4.0:
https://github.com/ansible/ansible/issues/31087

---

I would like to add this back later so we can kind of use https locally.  But I think it's best to wait until ansible to fixed.

### DIFF
--- a/lead_frontend.yml
+++ b/lead_frontend.yml
@@ -8,22 +8,22 @@
   # NOOP only to acquire IP address information about hosts.
   tasks: []
 
-- name: generate a self-signed certificate
-  hosts: 127.0.0.1
-  connection: local
-  vars:
-    cert_password: "password"
-    cert_domain: "*.{{ frontend_domain }}"
-    cert_dir: "{{ inventory_dir }}/secrets/certs"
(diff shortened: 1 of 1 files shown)


pumazi:
This can be transformed into a local_action anyhow, no?

karenc:
@pumazi Thanks!  I have a follow up pull request #390 that uses `local_action`.

PR #392, opened 2017-10-03
Update push_data to stop and start all publishing services

karenc:
Channel processing and publishing celery workers are connected to the
database so they need to be stopped before dropping the database.

Depends on #384.


karenc:
#384 has been merged so this has been unblocked.

PR #395, opened 2017-10-05
Upgrade pyramid_sawing to 1.1.3

karenc:
- Adjust the transit logging tween to set and return the response
  attribute of the request object.

This should fix the problem of getting "200 OK" for all requests in our
archive and publishing logs.


reedstrm:
Yay! I've been confused by this, for sure (tracebacks and 200s for the same request!)

PR #399, opened 2017-10-12
Add graylog settings to the qa variables

karenc:
We have set up syslog to forward logs to graylog if `graylog_server` and
`graylog_port` are defined.  So we just need to add the settings to the
qa environment group vars.


karenc:
[Unified logging](https://trello.com/c/re87Xkf5/159-unified-logging)

karenc:
We've set up an endpoint on graylog.cnx.org for qa.cnx.org. :+1:

PR #401, opened 2017-10-13
Change "include" to "import_tasks" or "import_playbook"

karenc:
The use of `include:` is deprecated in ansible v2.4.0 and will be
removed in v2.8:

```
[DEPRECATION WARNING]: 'include' for playbook includes. You should use 'import_playbook' instead. This feature will be removed in version 2.8. Deprecation warnings can be
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This feature
will be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is discouraged. The module documentation details page may explain more about this rationale..
This feature will be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

`include:` was used for dynamic and static reusable content, this has
now been separated into `import*` for static and `include*` for dynamic.

(The following is copied from
http://docs.ansible.com/ansible/latest/playbooks_reuse.html)

Differences Between Static and Dynamic

The two modes of operation are pretty simple:

 - Ansible pre-processes all static imports during Playbook parsing time.
 - Dynamic includes are processed during runtime at the point in which that
   task is encountered.

When it comes to Ansible task options like `tags` and conditional statements
(`when`:):

 - For static imports, the parent task options will be copied to all child
   tasks contained within the import.
 - For dynamic includes, the task options will only apply to the dynamic task
   as it is evaluated, and will not be copied to child tasks.

### DIFF
--- a/configure_hosts.yml
+++ b/configure_hosts.yml
@@ -1,7 +1,7 @@
 ---
 # Configure /etc/hosts file for development.
 
-- include: pretask.yml
+- import_playbook: pretask.yml
 
 - name: "configure /etc/hosts file"
   hosts:
--- a/copy_database.yml
+++ b/copy_database.yml
@@ -42,7 +42,7 @@
     dbmigrate_context_args: "{{ migration_contexts|default('--context cnx-publishing --context cnx-archive') }}"
     db_connection_string: "dbname={{ db_name }} user=postgres host=localhost port=5432"
   tasks:
-    - include: tasks/create_db.yml
+    - import_tasks: tasks/create_db.yml
       vars:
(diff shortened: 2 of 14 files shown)


pumazi:
Wow, ansible is getting its act together! These are great adjustments.

PR #406, opened 2017-10-17
add environments for each of the content servers

dtwilliamson:


dtwilliamson:
Oops, thanks for catching that!!

PR #408, opened 2017-10-18
Forward /static/ to publishing in varnish.vcl

karenc:
Publishing has static files like css and javascript files that are
currently 404 not found because varnish does not have routes for
`/static/`.

Add `req.url ~ "^/static/"` to varnish.vcl and use publishing as the
backend.

### DIFF
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -167,7 +167,7 @@ sub vcl_recv {
 
 {% endif %}
     # cnx rewrite archive
-    if (req.url ~ "^/a/") {
+    if (req.url ~ "^/a/" || req.url ~ "^/static/") {
         set req.backend_hint = publishing_cluster.backend(req.http.cookie);
         return (pass);
     }


karenc:
[How do we improve monitoring status of content production?](https://trello.com/c/lAYdPZi7/242-how-do-we-improve-monitoring-status-of-content-production)

reedstrm:
Part of me wants to move all the user-face pages into webview …

karenc:
They are admin pages, so I think it's ok for the pages to be in publishing.  We can also inline all the css...?  So this wouldn't be necessary.

reedstrm:
I guess my only real objection then is `/static/` is pretty generic - it'd make merging under openstax.org someday pretty painful, for example. Can we do `/a/static` instead? (I realize we'd need to do that over in cnx-publishing as well, then)

karenc:
Yeah that's what I thought as well.  `/a/` should be good, I'll change it in cnx-publishing as well.

pumazi:
@reedstrm, please make a story or two for the topics of your comments. This appears to be a bug fix rather than a feature. Thus I'm accepting it as a bug fix to make things work.

karenc:
Since we're moving `/static` to `/a/static` (see Connexions/cnx-publishing#199), this PR is no longer necessary.

PR #412, opened 2017-10-20
Filter out all haproxy health check from archive and publishing logs

karenc:
Archive and publishing rsyslog config filters out haproxy health check
requests from `lead_frontend`, but there are other servers with haproxy
installed.  The graylog stream for staging is showing health checks from
frontend (nginx) and legacy frontend (haproxy), so we need to exclude
these requests as well.

### DIFF
--- a/archive.yml
+++ b/archive.yml
@@ -1,5 +1,16 @@
 ---
 
+- name: collect info on frontend hosts
+  hosts:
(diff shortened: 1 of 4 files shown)


karenc:
[Unified logging](https://trello.com/c/re87Xkf5/159-unified-logging)

PR #413, opened 2017-10-23
Change requirements.txt's to allow branch or tag names

karenc:
Currently there is no way to specify the versions of a lot of the
backend packages installed in a dev environment.  The only way is to
modify the requirements.txt for archive, authoring or publishing
locally.

This commit makes it possible to provide git branch or tag names or
commit hash for different backend packages that are installed using git.

`rundeck.sh` is created for deploying to QA using rundeck.


rundeck.sh
```diff
@@ -0,0 +1,42 @@
...
+extra_vars+="cnx_authoring_version='${CNX_AUTHORING_VERSION}' "
+extra_vars+="cssselect2_version='${CSSSELECT2_VERSION}' "
+extra_vars+="cnx_easybake_version='${CNX_EASYBAKE_VERSION}' "
+extra_vars+="cnx_publishing_version='${CNX_PUBLISHING_VERSION}' "
+extra_vars+="cnx_recipes_version='${CNX_RECIPES_VERSION}' "
+extra_vars+="webview_version='${WEBVIEW_VERSION}'"
+
+set -x
+cd /var/cnx-deploy
+ansible-playbook -i "environments/qa/inventory" \
```
reedstrm: The PR talks about dev environment, o.k. I can see the use-case for matching branches for dev work. The code here is changing QA - well, yes, I guess if we're doing the qa-on-branch model (which I'm still not 100% on board with, but …) Let's just be sure two things happen - 1. the set of what-was-deployed is recorded (I guess that's what all the work to get a /revision.txt and whatever is about) and 2. that this never propagates up to staging or production.

karenc:
This PR only changes `environments/__dev_envs/files`, you can see clearly that it is very different from `environments/staging/files` and `environments/prod/files`.  We can't possibly use this for staging and prod.  So there's no need to worry about it.

karenc:
This is meant for rundeck, the UI will just have a long list of input boxes and whoever triggering the deployment should know which repos need which branch for testing.  We usually include that information in PRs, so people should know.

reedstrm:
Right, I realized this does not provide a mechanism to change staging/prod. Just wanted to forestall any "can't we just do that on staging too?" requests in the future. :-) Yes, that's about what I expected for the rundeck UI. Just thinking about ways to make it even easier, perhaps  - one input box that you type the feature-branch-name into, and if that name exists for each repo, that is what is checked out, otherwise you get HEAD or most-recent-tag, depending on environment.

karenc:
I would say that's an enhancement :laughing:  It's good to think about it.  Can definitely do it in the future.

A bigger problem that we need to think about is, what happens if the cnx-db branch has migrations.....  I guess it might require some manual rolling back etc... :fingers_crossed:  Hopefully it doesn't happen that often.

reedstrm:
Ohh, good point - if we do migrations-on-branch, we need to be able to restore QA to a "snapshot" state, or rollback, right - that code needs testing, too!

philschatz:
From discussing, it sounds like this is missing UX from Rundeck.

pumazi:
This is still relevant, but it may now be going in a direction different from what we are hoping to do for openstax/cnx#292. Due to that I'm going to close the PR.

PR #417, opened 2017-10-31
Add cnx robots.txt to /var/www/webview

karenc:
We had robots.txt in cnx-archive which was not really the right place to
put it.  DevOps prefer to have robots.txt in cnx-deploy so they can
configure it.

### DIFF
--- a/roles/robots_txt/vars/main.yml
+++ b/roles/robots_txt/vars/main.yml
@@ -1,0 +1,24 @@
+---
+robots_txt:
+  user_agents:
+    - {name: 'ScoutJet', crawl_delay: 10}
+    - {name: '*'}
+    - {name: 'Baiduspider', crawl_delay: 10}
+    - {name: 'BecomeBot', crawl_delay: 20}
+    - {name: 'Slurp', crawl_delay: 10}
+  disallow_list:
+    - legacy.cnx.org
+    - /lenses
+    - /browse_content
+    - /content/
+    - /content$
+    - /*/pdf$
+    - /*/epub$
+    - /*/complete$
+    - /*/offline$
+    - /*?format=*$
+    - /*/multimedia$
+    - /*/lens_add?*$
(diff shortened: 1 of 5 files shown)


roles/robots_txt/vars/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    - /*/pdf$
+    - /*/epub$
+    - /*/complete$
+    - /*/offline$
+    - /*?format=*$
+    - /*/multimedia$
+    - /*/lens_add?*$
+    - /lens_add
+    - /*/lens_view/*$
+    - /content/*view_mode=statistics$
```
pumazi: I'm not a fan of buried configuration. But maybe I'm looking at this stuff wrong.

karenc: You mean you would prefer to move this to group vars?  I can also move this to the template (but move the user agent config to group vars)

pumazi:
We should get this into place soon'ish since the current robots.txt points to `/sitemap.xml`, which is an invalid url. I'd create a separate issue, but I don't want to muddy things more than they might already be (that is, multiple input steams to the devops team). So let's just try to get this pushed through.

roles/robots_txt/vars/main.yml
reedstrm: Hmm, there's actually two levels of config here - user-agents to disallow would be universal, but which paths are not-of-interest for crawling is an aspect of the web-apps deployed, so shouldn't those apps configure fragments for robots.txt, like the haproxy config does?

reedstrm: So, I think I'm voting "vars" right?

karenc: These are all legacy urls, right?

reedstrm: yes

PR #432, opened 2017-11-20
Remove postgres superuser privileges from db user

karenc:
When initializing the schema, there are some functions that require
postgres superuser privileges to create.  But we don't really want the
db user to have superuser privileges.

Create a temporary super user for `cnx-db init` and `cnx-db venv` and
then reassign everything to the db user.

### DIFF
--- a/database.yml
+++ b/database.yml
@@ -26,14 +26,6 @@
       vars:
         db_user: "{{ archive_db_user }}"
         db_password: "{{ archive_db_password }}"
-        # FIXME (8-Mar-12017) db-role-perms:
-        #       New databases and migrations currently
-        #       require the user to have superuser privileges.
-        #       This is only required for database
-        #       initialization and migrations.
-        #       This is to be correct in cnx-db and db-migrator.
-        #       In the mean time use xxx_archive_db_user_role_attr_flags.
-        role_attr_flags: "{{ xxx_archive_db_user_role_attr_flags|default('CREATEDB,NOSUPERUSER') }}"
     - import_tasks: tasks/create_db.yml
       vars:
         db_owner: "{{ archive_db_user }}"
--- a/roles/repository_db/handlers/main.yml
+++ b/roles/repository_db/handlers/main.yml
@@ -3,22 +3,45 @@
 # Note, the setup current **assumes** that the publishing and archive database
 # is the same database. There has been no reason to split these yet, so we won't.
 
+- name: initialize repository database (create superuser)
+  become: yes
+  become_user: postgres
+  postgresql_user:
+    db: postgres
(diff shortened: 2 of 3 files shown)


database.yml
```diff
@@ -26,14 +26,6 @@
...
         db_user: "{{ archive_db_user }}"
         db_password: "{{ archive_db_password }}"
-        # FIXME (8-Mar-12017) db-role-perms:
-        #       New databases and migrations currently
-        #       require the user to have superuser privileges.
-        #       This is only required for database
-        #       initialization and migrations.
-        #       This is to be correct in cnx-db and db-migrator.
-        #       In the mean time use xxx_archive_db_user_role_attr_flags.
-        role_attr_flags: "{{ xxx_archive_db_user_role_attr_flags|default('CREATEDB,NOSUPERUSER') }}"
```
pumazi: 💯  for the removal of a FIXME

roles/repository_db/handlers/main.yml
```diff
@@ -3,22 +3,45 @@
...
-- name: initialize repository database (dbmigrator init)
+- name: initialize repository database (reassign owner)
   environment:
+    # If PGHOST is not set, psql tries to look for cluster "main" which doesn't
+    # work:
+    #     $ psql -U postgres -l
+    #     Error: Invalid data directory
+    # This appears to be a problem after a postgresql-common update
+    PGHOST: "{{ archive_db_host }}"
+  command: "psql -U postgres {{ archive_db_name }} -c 'REASSIGN OWNED BY super_{{ archive_db_user }} TO {{ archive_db_user }}; DROP USER super_{{ archive_db_user }};'"
```
pumazi: Why are you doing a `DROP USER ...` here?

pumazi: I'm anticipating we'll want this user around for select migrations.

karenc: Oh... `dbmigrator migrate` uses the `postgres` user by default.  That's what we're using at the moment.  May be we should also change dbmigrator to use `DB_URL` and `DB_SUPER_URL`?

pumazi: 🤔 I guess that makes sense. I've been thinking of the `postgres` role as if it were synonymous with `root`. And that you wouldn't want to use the `root` user unless you had to, but I'm pretty sure that's just a misunderstanding on my part, because the postgres role is just a superuser role.

So this works for me. But why not use the postgres user for db init here?

karenc: That's because `REASSIGN OWNER` cannot be used with the postgres user :p

PR #434, opened 2017-11-22
Use delegated facts instead of NOOP to gather facts

karenc:
We used `tasks: []` to gather facts (mostly ip addresses) from other
servers.  There is a different way of doing this in ansible 2 called
delegated facts.

http://docs.ansible.com/ansible/latest/playbooks_delegation.html#delegated-facts

This allows us to gather facts from other hosts that we need for
configuration.

### DIFF
--- a/archive.yml
+++ b/archive.yml
@@ -1,16 +1,5 @@
 ---
 
-- name: collect info on frontend hosts
-  hosts:
-    - lead_frontend
-    - frontend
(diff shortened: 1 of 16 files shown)


reedstrm:
Oh, this looks much nicer - makes `--limit` much more likely to work correctly, without lots of circumlocution in the playbook.

PR #436, opened 2017-11-28
Allow cache purges from any connected hosts

### LINKED ISSUE #129: Allow purging from any connected host
## Summary

Allow any environment host to make HTTP `PURGE` requests to varnish. Currently we only allow requests made locally to use the `PURGE` method. Because cnx-publishing has the ability to purge on publish/bake, we need to allow, at the very list, the server hosting cnx-publishing to make `PURGE` requests to varnish(s).

Documentation for where to allow this can be found at [Purge and banning](https://www.varnish-cache.org/docs/4.0/users-guide/purging.html).
## Environment

all

karenc:
Include all the IP addresses from all the hosts except the frontend ones
(the ones with varnish) in the acl purge list.

If the frontend hosts are included, then any host can issue a PURGE
request which may not be what we want.

Close #129

### DIFF
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -107,14 +107,8 @@ backend {{ backend_name }} {
 acl purge {
     "localhost";
     "127.0.0.1";
-    "{{ frontend_domain }}";
-{% set frontend_ips = groups.frontend|map('extract', hostvars, ['ansible_default_ipv4', 'address'])|list %}
-{% for host in groups.all %}
-{% if hostvars[host].ansible_default_ipv4.address not in frontend_ips %}
-{# we have to exclude the frontend ip address because otherwise
-   "curl -X PURGE https://cnx.org/" would work from any hosts #}
+{% for host in groups.varnish_purge_allowed %}
     "{{ hostvars[host].ansible_default_ipv4.address }}";
-{% endif %}
 {% endfor %}
 }
 
@@ -318,14 +312,24 @@ sub vcl_recv {
     }
     
     if (req.method == "PURGE") {
-        if (!client.ip ~ purge) {
+        # Change varnish to use `req.http.x-forwarded-for` instead of `client.ip`
+        # because `client.ip` has the IP address of haproxy (I think).
+        # `req.http-x-forwarded-for` sometimes has multiple IP addresses because
+        # of redirects within our servers, the first IP is what we want, so remove
+        # everything after ",".
+        if (!std.ip(regsub(req.http.x-forwarded-for, ",.*$", ""), "0.0.0.0") ~ purge) {
             return (synth(405, client.ip));
         }
         ban("req.url ~ " + req.url + "$");
         return (synth(200, "Ban added"));
     }
    if (req.method == "PURGE_REGEXP") {
-        if (!client.ip ~ purge) {
+        # Change varnish to use `req.http.x-forwarded-for` instead of `client.ip`
+        # because `client.ip` has the IP address of haproxy (I think).
+        # `req.http-x-forwarded-for` sometimes has multiple IP addresses because
+        # of redirects within our servers, the first IP is what we want, so remove
+        # everything after ",".
+        if (!std.ip(regsub(req.http.x-forwarded-for, ",.*$", ""), "0.0.0.0") ~ purge) {
             return (synth(405, "Not allowed."));
         }
         ban("req.url ~ " + req.url);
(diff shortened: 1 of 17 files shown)


karenc:
Sorry I accidentally pushed to master...  :(

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -108,7 +108,14 @@ acl purge {
     "localhost";
     "127.0.0.1";
     "{{ frontend_domain }}";
-    # TODO Include IP addresses of other serving hosts.
+{% set frontend_ips = groups.frontend|map('extract', hostvars, ['ansible_default_ipv4', 'address'])|list %}
+{% for host in groups.all %}
+{% if hostvars[host].ansible_default_ipv4.address not in frontend_ips %}
+{# we have to exclude the frontend ip address because otherwise
+   "curl -X PURGE https://cnx.org/" would work from any hosts #}
+    "{{ hostvars[host].ansible_default_ipv4.address }}";
```
reedstrm: Ah, because of the 443 -> 80 rewrite. We could fix that in the test by looking if there is an `x-forwarded-for` header, I suppose … But this works, at the risk of making purging hard on single-machine-deploys, no?

karenc: It should work as it'll match `localhost`.

karenc:
I think this is still white-listing.

reedstrm:
Your code uses the all group, then removes the frontends. Which makes it a potential surprise if someone in the future adds another type of host group - they get granted purge access by default. Not a huge deal, but was wondering about the thinking.

roles/varnish/templates/etc/varnish/varnish.vcl
reedstrm: Not sure this is true - doesn't your local setup have `frontend_domain` aliased to `localhost` on /etc/hosts? Just tested locally myself:
```reedstrm@sensei:~/src/rewrite/cnx-publishing (direct-results-access $%)$ python -m SimpleHTTPServer
Serving HTTP on 0.0.0.0 port 8000 ...
10.64.71.229 - - [29/Nov/2017 11:11:22] "GET / HTTP/1.1" 200 -
10.64.71.229 - - [29/Nov/2017 11:11:23] code 404, message File not found
10.64.71.229 - - [29/Nov/2017 11:11:23] "GET /favicon.ico HTTP/1.1" 404 -
127.0.0.1 - - [29/Nov/2017 11:11:37] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [29/Nov/2017 11:11:37] code 404, message File not found
127.0.0.1 - - [29/Nov/2017 11:11:37] "GET /favicon.ico HTTP/1.1" 404 -
```
That's two successive hits on `http://10.64.71.229:8000` and `http://localhost:8000`
Again not a huge deal, it means to purge from the frontend, you have to use localhost

karenc:
Ah I see.  We can change it to `legacy_frontend`, `archive` and `publishing` instead.  I think we still need to exclude the frontend IP in case any of the group hosts match.  But really, only `publishing` need to purge I guess?

reedstrm:
Yeah, that should work - legacy + publishing. Agreed we don't want to allow arbitrary purges - had that happen once w/ someone spider that wanted 'fresh' results. Ouch.

reedstrm:
P.S. I love the one-line python http server!

karenc:
Yes, I have `127.0.1.1 local.cnx.org` in my `/etc/hosts` file.  I see the same thing on staging02.  But may be it's not a standard thing... I don't know.

reedstrm:
Ah yup:

karenc:
Now that I look at this... I'm wondering how multi machine deploys work... I'm going to test this some more locally.

karenc:
I've changed to whitelisting again with a new group `varnish_purge_allowed` so it can be configured from the inventory file. I did a simplified version of multi machine deploy and it appeared to work.  Single machine deploy worked for me too.

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -318,14 +312,14 @@ sub vcl_recv {
     }
     
     if (req.method == "PURGE") {
-        if (!client.ip ~ purge) {
+        if (!std.ip(regsub(req.http.x-forwarded-for, ",.*$", ""), "0.0.0.0") ~ purge) {
```
reedstrm: took me a second to understand this, but I _think_ it makes sense now …

karenc: I'll add some comments

PR #441, opened 2017-12-04
Enable sysstat on all servers

### LINKED ISSUE #437: Enable sysstat on all servers
## Summary

We may need more information on the servers when debugging performance problems.

## Environment

all

karenc:
Install sysstat and enable it on all servers so we can collect stats
from the servers and help debug performance problems.

Close #437

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -67,6 +67,10 @@
     - iptables
     - firewall
 
+- import_playbook: sysstat.yml
+  tags:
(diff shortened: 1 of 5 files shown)


reedstrm:
note bene - @dtwilliamson is not available until tomorrow, but today is code freeze for next release, so I'm jumping in w/ some PR review.

PR #442, opened 2017-12-04
Update pyup.yml to update requirements.yml asap

karenc:
According to https://pyup.io/docs/bot/config/#schedules:

By default, the bot will open a Pull Request as soon as there's an
update for one of your dependencies available. If this creates too much
noise, you can tell the bot to hold back updates and bundle them in a
single Pull Request on a daily, weekly, biweekly or monthly basis.

Which is what we want.  So we don't need to specify a schedule for pyup
bot.


pumazi:
I'm glad we've gotten to the point we'll be able to do on-demand updates.

PR #451, opened 2017-12-07
Run deferred migrations at the end of main.yml

karenc:
Instead of having to remember to run deferred migrations, we're going to
run deferred migrations at the end of every deployment.  It will run on
one of the publishing hosts, the output will be in
/tmp/deferred-migrations.out and /tmp/deferred-migrations.err.  If
deferred migrations are already running, we'll skip this step.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -69,6 +69,8 @@
 
 - import_playbook: update_versions.yml
 
+- import_playbook: run_deferred_migrations.yml
+
 - name: "notify #cnx-stream of the deployment (end)"
   hosts: all
   vars:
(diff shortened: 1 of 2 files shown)


reedstrm:
My biggest concern with this is for production rollouts - when does service restart happen relative to the migrations occurring? Is the restarting service going to see a half-migrated state, potentially? Also, devops needs to be able to know very well when deployment of new code is complete, and that the system is live, as well as knowing that deferred migrations are running, and guesstimates of when they might end. Hopefully tailing the files in /tmp will be enough.

karenc:
The restarting of services were always done in cnx-deploy.  The deferred migrations were run by hand afterwards, I'm not sure what your concern is exactly.  Code should not depend on the deferred migrations  and services are not restarted after running deferred migrations.

reedstrm:
I talked w/ @dtwilliamson about this, and he agrees with you all, so I'm good. Merging.

PR #452, opened 2017-12-07
Always restart archive and publishing

karenc:
The current logic to restart archive and publishing when code or
configuration has changed is not wrong.  The problem is ansible doesn't
restart archive and publishing immediately when we notify the handlers,
it only does it after all the tasks in the same playbook are finished.
If any task fails in the middle, the services are not restarted.  The
next time we run cnx-deploy, it'll see that the code and configuration
haven't changed and so does not try to restart the services.  This
happened so many times that may be we will just always restart the
services.

### DIFF
--- a/roles/archive/tasks/main.yml
+++ b/roles/archive/tasks/main.yml
@@ -96,6 +96,9 @@
   pip:
     requirements: "/var/lib/cnx/archive-requirements.txt"
     virtualenv: "/var/cnx/venvs/archive"
+
+- name: restart archive
+  command: "/bin/true"
   notify:
     - restart archive
 
--- a/roles/publishing_common/tasks/main.yml
+++ b/roles/publishing_common/tasks/main.yml
@@ -100,6 +100,9 @@
   pip:
     requirements: "/var/lib/cnx/publishing-requirements.txt"
     virtualenv: "/var/cnx/venvs/publishing"
+
+- name: restart publishing
+  command: "/bin/true"
   notify:
     - list supervisor applications
     - restart publishing


karenc:
@pumazi If there's a different way of detecting the services didn't restart when we try to run cnx-deploy again, then we can do that instead.

reedstrm:
Discussed this w/ @dtwilliamson  on Thursday - he's has concerns re: interaction between restarting and errors.

karenc:
Could you explain more about the concerns @dtwilliamson has?

dtwilliamson:
I just participated in the discussion, that's all. My preference is to not restart when it's not necessary - but it's not a hard and fast requirement. @reedstrm and @Dantemss and I were discussing some way to get ansible to trigger handlers immediately (which might make them get run multiple times - not good) or to have them run even if there are fatal steps. @reedstrm can probably tell you what the conclusion was.

karenc:
Oh, of course that would be the best solution.  This is a workaround.  Is @Dantemss planning to submit a pull request to ansible to get this to work?

reedstrm:
Hmm, the general problem being addressed here is the observation that services don't always restart, even if they've had new cod deployed. One possible path to that has to do with handlers not running because other parts of  playbook error, after new code or config has deployed to a service. Then on re-run, you no longer have modified that particular service, so it is not restarted, leading to new code deployed, but service running old code. This change would cause every run to see the services as changed, so would notify the restart handler. Seems ok for out fairly fast-starting, light weight services.

However, I'm not sure we've actually _seen_ the path above - I don't see errors and restarts in the cnx-deploy logs, bug sometimes the services don't restart. This will probably fix that case as well.

karenc:
A few months ago, we were not restarting services unless there was a configuration change, but that should work now.  What usually happens is with publishing, we run migrations after updating the code, and if migrations fail (which happens quite often), publishing doesn't get restarted.  I think archive doesn't have this problem as often.

Dantemss:
The quick and easy workaround is to add force_handlers: True to the play. That will trigger the handlers to run even if the play fails. However, it will not trigger the handlers to run if you ctrl + c out of it or the server runs out of memory or the network drops.

A more elaborate workaround that I've been thinking about is adding a task that will `stat` some file and save that in `some_var`, then another task (that runs after the above) that creates said file and finally a handler that deletes the file. In all tasks that notify handlers, add this: `changed_when: "{{ True if some_var else omit }}"`. And at the end of the play add a task to notify the handler that deletes the file (or figure out how to make it run after the other handlers). This will make it so if the play is aborted, then ALL handlers will be notified next time the play runs.

karenc:
Thanks @Dantemss, that's really useful stuff! We've never run out of memory. I think force handlers will work for us.

Dantemss:
Well @dtwilliamson ended up convincing me that this would be way too much code for a workaround, so I think we'll end up just making sure our handlers in tutor-deployment always run using `changed_when: True` on the notifying tasks... The handlers don't even take that long anyway.

Dantemss:
OK this is what I ended up with in `tutor-deployment`:

`force_handlers = True` in `ansible.cfg` will usually get the handlers to run, even on failures

For those cases that `force_handlers` doesn't cover, I added tasks to each role that has a handler that look like this:

```
- name: notify handler
  command: 'true'
  notify: fail2ban
  when: notify_all_handlers
  tags:
    - fail2ban
    - handler
```

(inspired by your code that uses `command: 'true'`, much better than the suggested way of using the debug module)

and notify_all_handlers is in group_vars/all:

`notify_all_handlers: "{{ notify_all | d | bool }}"`

So you can force all handlers to run like this:

`ansible-playbook ... -e notify_all=True`

and you can run only the handlers:

`ansible-playbook ... --tags handler -e notify_all=True`

(my handlers also have the handler tag... not sure if that matters)

PR #455, opened 2017-12-13
(DISCUSS) Change haproxy to use leastconn for cnx.org varnish

karenc:
We were using uri hashing for varnish before.  The reason is we have
multiple varnish instances, if we use uri hashing, there is a higher
chance of getting a cache hit.

I did some experiment with different load balancing algorithms.  I set
up 2 varnish instances and 2 archive instances, one has `time.sleep(5)`
added to it to show what happens if archive is being slow.  I have 500
archive urls and 10 wget processes that fetch 50 urls each.

1. Using uri hashing:

   Fetching all the urls with no cache took 57 seconds.  The normal
   archive served 457 urls and the slow one served 42.

   Fetching all the urls again took 6 seconds.  All the subsequent
   fetches took less than 2 seconds.

2. Using leastconn:

   Fetching all the urls with no cache took 53 seconds.  The normal
   archive served 459 urls and the slow one served 40.

   Fetching all the urls the second time took 51 seconds, the third time
   45 seconds, the fourth time 50 seconds.

3. Using roundrobin with maxconn 4:

   Fetching all the urls with no cache took 42 seconds.  The normal
   archive served 470 urls and the slow one served 30.

   Fetching all the urls the second time took 51 seconds, the third time
   1 minute 5 seconds, the fourth time 25 seconds.

Archive instances being slow may not be the problem we saw on
production.  Perhaps we were having problems with varnish instead.  In
that case, using leastconn or roundrobin might give us better
performance overall.

### DIFF
--- a/roles/lead_load_balancer/tasks/main.yml
+++ b/roles/lead_load_balancer/tasks/main.yml
@@ -43,7 +43,7 @@
 
       # Load balancing the frontend grouping
       backend nodes
-        balance uri
+        balance leastconn
         option httpchk GET /ping HTTP/1.1\r\nHOST:\ {{ frontend_domain }}
         option forwardfor
 


dtwilliamson:
Maybe I'm not looking at this correctly, but it seems to me that round robin had the best results in your comparison. Also the branch name and the title of this pull request don't match so I'm additionally confused.

karenc:
Ah yeah... when I created the branch, I was going to use roundrobin, but then leastconn seems to be about the same.

I would say the current uri hashing has the best performance, if you look at subsequent fetches.  It's down to 1 - 2 seconds for 500 urls by the third fetch.

dtwilliamson:
Oops, I meant to say uri hashing (corrected).

reedstrm:
The observed behavior was when one archive host got slow for some reason (I think this may have been the leaking DB connection problem?), w/ uri hashing it didn't fallback to the other server, at all. We even had errors happening, even though the other server was not loaded at all. 

I pointed out that in addition to these errors, response times from archive seem to have dropped from when we had the hand-rolled install. Old config was varnish -> haproxy ->  archive[0-9], using leastconn.

Now we have haproxy -> varnish[01] -> archive[00-10], using uri-hash to select which varnish instance - varnish uses round_robin to select among archive instances. (At first I thought varnish only talked to local archive instances, which made uri-hash potentially dangerous, but I since looked more closely at the deployed config, and realized we have both caches talking to all 10 backends)

My concern with this is that haproxy is optimized for load balancing - it has one job, and does it very well (I bodged some site-redirection in there as well  - files stuff - but it was non-critical path) … Varnish is using a director to select which archive to use, via round robin, with a rather simplistic health heck. All this code is fairly recent in the history of varnish, and is a bit of a "me too" feature set, IMHO. Varnish's "one job" is (and should) be caching.

Measuring things is a good idea. URI hashing seems to be a win, in non-error cases. How it responds to degradation of backends (beyond slow response) is an open question.

karenc:
This is not a non-error case exactly, I had one archive with a 5 second delay before returning anything.  You can see that the slow archive is getting a lot less requests (whether we use uri hashing, leastconn or roundrobin), so actually it's working fine.

karenc:
So the reason why I said may be varnish was the problem is remember we had problems with prod02.  Varnish is on prod01 and prod02, and I think prod02 was being slow... But it was considered healthy by haproxy.  Does anyone else remember something like that?

reedstrm:
Yup, prod02 was being slow to the point of failing. We got timeouts. I don't think we ever solved what the exact problem was.

karenc:
Yeah, is that enough reason to use leastconn / roundrobin...? :thinking:

reedstrm:
(actually the files stuff is over in nginx, where we also do the SSL termination)

reedstrm:
Perhaps what we really need is some additional tuning of the backend probes in varnish. Default timeout is 2s !! https://varnish-cache.org/docs/4.1/reference/vcl.html#reference-vcl-probes

karenc:
Yes I looked at that too, the varnish probe was pinging archive every 2-3 seconds.  I think we should probably change it to 30 seconds / 1 minute, pinging `/extras` which uses the database instead of `/` which is just 404.

edwoodward:
Closing because this will not be merged

PR #456, opened 2017-12-13
Unified file in collections

### LINKED ISSUE #439: Need to add unified file for collection type
## Summary

In order to use the folder_contents hack to add a ruleset.css recipe to a collection, legacy needs
a setup tweak to add UnifiedFile as an allowed child content type of Collection in portal_types.

On production this has been set manually TTW via admin access to /portal_types/manage_main

(/portal_types/Collection/manage_propertiesForm)

## Environment
content01 on ubuntu 16.04 (but all legacy deploys need this)

karenc:
- Include UnifiedFile in Collection allowed_content_types

  In order to add a ruleset.css recipe to a collection, `UnifiedFile`
  needs to be added as a allowed content type in the ZMI.
  
  I was looking into doing this in Products.RhaptosCollection
  /profiles/default/types/Collection.xml where the type is
  defined, but running buildout doesn't seem to pick the changes up.  May
  be we're not reinstalling the products every time we run buildout.
  
  Close #439

- Combine rhaptos site set up tasks into one file

  Merge `tasks/set_up_rhaptos_print_host.yml` and
  `tasks/set_up_localfs_folders.yml` into
  `tasks/set_up_rhaptos_print_host.yml`.  There is only one task in each
  of the task files so can just put them in the same file.

### DIFF
--- a/tasks/set_up_rhaptos_site.yml
+++ b/tasks/set_up_rhaptos_site.yml
@@ -4,3 +4,8 @@
   shell: "./bin/instance run scripts/set_up_localfs_for_pdfs.py {% for item in archive_exports_directories|default(default_archive_exports_directories) %}pdfs{{ loop.index }}:/var/www/{{ item }} {% endfor %}"
   args:
     chdir: "/var/lib/cnx/cnx-buildout"
+
+- name: set up rhaptos print host
+  shell: "echo -c \"app.plone.rhaptos_print._host = 'http://{{ frontend_domain }}'; import transaction; transaction.commit()\" | ./bin/instance debug"
+  args:
+    chdir: "/var/lib/cnx/cnx-buildout"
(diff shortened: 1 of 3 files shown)


tasks/set_up_rhaptos_site.yml
```diff
@@ -0,0 +1,16 @@
...
+    chdir: "/var/lib/cnx/cnx-buildout"
+
+- name: set up rhaptos print host
+  shell: "echo -c \"app.plone.rhaptos_print._host = 'http://{{ frontend_domain }}'; import transaction; transaction.commit()\" | ./bin/instance debug"
+  args:
+    chdir: "/var/lib/cnx/cnx-buildout"
+
+- name: set up collections allowed content types
+  shell: "echo -c \"app.plone.portal_types.Collection.allowed_content_types += ('UnifiedFile',); import transaction; transaction.commit()\" | ./bin/instance debug"
+  args:
```
reedstrm: Not idempotent, so we _might_ end up w/ duplicates, but this is only run on initial setup, so probably not.
A fix would be to use something like:
```app.plone.portal_types.Collection.allowed_content_types += [('UnifiedFile',),()]['UnifiedFile' in app.plone.portal_types.Collection.allowed_content_types]```
(This uses the False==0, True==1 hack to index into a two-value list to only make the += a noop if the value is already present)

PR #457, opened 2017-12-13
Remove hourly archive restart cron job

karenc:
We removed daily and weekly archive restart cron jobs but we have hourly
cron jobs on production, so need to remove that too.


reedstrm:
Are we sure we've solved the hanging db connection problem, before we deploy this?

karenc:
We already tried to remove the cron jobs in #411.  I just somehow removed the daily and weekly ones but not hourly ones...

PR #463, opened 2017-12-15
Move cnx logs from /var/log/syslog to /var/log/cnx.log

### LINKED ISSUE #458: Application logs should not go in /var/log/syslog
Changes need to be made to `/etc/rsyslog.d/50-default.conf` so that user and local level log entries don't go in `/var/log/syslog`.

karenc:
- Rotate /var/log/cnx.log every week

  /var/log/cnx.log has what we send to graylog, to keep the files from
  growing too big, we are rotating the files every week.  The cnx
  logrotate config file is based on the rsyslog logrotate config file.

- Move cnx logs from /var/log/syslog to /var/log/cnx.log

  There is no need to have the cnx logs to be in /var/log/syslog to get
  them forwarded to graylog.  It is good to have /var/log/cnx.log which
  has everything that should be in graylog.
  
  Close #458

### DIFF
--- a/roles/archive/meta/main.yml
+++ b/roles/archive/meta/main.yml
@@ -6,3 +6,4 @@ dependencies:
   - memcached
   - supervisord
   - rsyslog
+  - logrotate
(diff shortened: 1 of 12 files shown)


reedstrm:
The fun bits are filters written in: http://www.rsyslog.com/doc/v8-stable/configuration/index.html

PR #469, opened 2017-12-20
Include collection.xml in prod db slim dump

### LINKED ISSUE #466: Slim dump collection.xml files contain 'dummy file'
## Summary

I was attempting to do an xpath query on collection files using the slim dump data. But it appears that our slim dump logic paves over 'collection.xml' named files.

```
ERROR:  invalid XML document
DETAIL:  line 1: Start tag expected, '<' not found
dummy file
^
```

## Environment

N/A

karenc:
I didn't finish testing this..  but it should work :p

Close #466

### DIFF
--- a/acquire_data.yml
+++ b/acquire_data.yml
@@ -24,6 +24,7 @@
       - "index_auto_generated.cnxml"
       - "ruleset.css"
       - "featured-cover.png"
+      - "collection.xml"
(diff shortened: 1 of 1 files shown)


pumazi:
Yeah, I don't see why it shouldn't work. I didn't realize this was a oneliner fix. Nice!

PR #472, opened 2018-01-03
Add backup role in database.yml

karenc:
- Add backup host to environments

  Set up backup in different environments and set slim dumps to yes for
  production.

- Add backup role in database.yml

  The backup role creates a weekly cnx database dump and optionally a slim
  database dump.

### DIFF
--- a/environments/prod/host_vars/backup2.cnx.org
+++ b/environments/prod/host_vars/backup2.cnx.org
@@ -1,0 +1,4 @@
+---
+slim_dumps: yes
+# Point at the replicant:
+archive_db_host: prod10.cnx.org
--- a/roles/backup/templates/etc/cron.weekly/cnx-slim-dump
+++ b/roles/backup/templates/etc/cron.weekly/cnx-slim-dump
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Create a slim dump of the cnx production database so developers can use it
+# for development and testing
+
+# 0. Make sure this server can access the production database (with pg_dump or
+#    ssh)
+
+# Target databases settings
+SLIM_DUMP_DB_NAME=cnx_slim_dump
+SLIM_DUMP_FILENAME="/var/backups/$SLIM_DUMP_DB_NAME.$(date +%Y-%m-%d).sql.gz"
+
+# Source databases settings
+ARCHIVE_DB_HOST={{ archive_db_host }}
+ARCHIVE_DB_PORT={{ archive_db_port }}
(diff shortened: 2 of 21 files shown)


roles/backup/templates/etc/cron.weekly/cnx-slim-dump
```diff
@@ -0,0 +1,73 @@
...
+    bname=$(basename $SLIM_DUMP_FILENAME)
+    ls -1 -r $(dirname $SLIM_DUMP_FILENAME)/${bname/.*/}* | sed -n '4,$ p' | xargs rm -f
+    psql -U postgres -c "DROP DATABASE IF EXISTS $SLIM_DUMP_DB_NAME;"
+}
+trap cleanup EXIT
+
+# 1. Create a temporary database for the slim dump
+psql -U postgres <<EOF
+DROP DATABASE IF EXISTS $SLIM_DUMP_DB_NAME;
+CREATE DATABASE $SLIM_DUMP_DB_NAME;
```
reedstrm: Pretty sure this will fire
```
ERROR:  cannot execute CREATE DATABASE in a read-only transaction
```
 if pointed at a replicant, so should _not_ be set for production.

environments/prod/host_vars/backup2.cnx.org
```diff
@@ -0,0 +1,4 @@
+---
+slim_dumps: yes
+# Point at the replicant:
+archive_db_host: prod10.cnx.org
```
reedstrm: slim dump script and replicant are not compatible.

roles/backup/templates/etc/cron.weekly/cnx-slim-dump
pumazi: This is using the local database on the backup server.

environments/prod/host_vars/backup2.cnx.org
pumazi: The backup script using the `archive_db_host` value here. The database dump to the backup server is being created from the database replicant.

reedstrm: Ah, I see now - it's a two step: backup uses `archive_db_host` to make a local copy via `pg_dump`, then slim dump is created locally on backup server.

roles/backup/templates/etc/cron.weekly/cnx-slim-dump
reedstrm: Which is obvious if I read closely - there's no hostname provided, so has to be happening locally.

PR #477, opened 2018-01-05
first cut diagram of deployed production cluster

reedstrm:
Drew an SVG to help me remember what 'prod05' has running on it, etc.

### DIFF
--- a/docs/deploy-prod.svg
+++ b/docs/deploy-prod.svg
@@ -769,7 +769,7 @@
      inkscape:pageopacity="0.0"
      inkscape:pageshadow="2"
      inkscape:zoom="1.979899"
-     inkscape:cx="297.91842"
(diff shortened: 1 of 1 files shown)


pumazi:
Why didn't you work from my existing diagrams? See those at https://github.com/openstax/napkin-notes/blob/master/pumazi/cnx/cnx-deploy-phases.md

Granted the diagram source wasn't made public. These are more detailed than what you have here. The source for these diagrams can be found in https://github.com/pumazi/diagrams and you can edit them over at draw.io (an open source diagramming tool).

I'd suggest updates be made to these diagrams I've made and the source and rendered results be stored here.

pumazi:
I've added the existing diagrams file in #478 as a followup to this. Please update that diagram using draw.io and upload a resulting PNG. Thank you.

PR #478, opened 2018-01-08
Add existing phased diagrams to the project documentation

pumazi:
This is a quick commit to include the existing diagrams that use the open source draw.io tool. You can see the results of the included `prod-diagram.xml` file at [Napkin Notes: cnx-deploy phases doc](https://github.com/openstax/napkin-notes/blob/master/pumazi/cnx/cnx-deploy-phases.md).

This commit is my followup to #477.


reedstrm:
I've researched a bit, and  I guess I'm o.k. w/ draw.io suite of tools for the source format, as long as we export SVGs of each drawing page as well, both as the display version, and as a potential edit source in the future.

reedstrm:
Did some editing, correcting the above issues w/ the production image. I'm not particularly enamored with
the look of the diagram, but it's better than nothing. Not sure it's significantly better than mine was, but meh.

PR #492, opened 2018-01-31
need to use archive virtualenv for venv setup

reedstrm:
- publishing is not deployed to db server

### DIFF
--- a/copy_database.yml
+++ b/copy_database.yml
@@ -82,7 +82,9 @@
       environment:
         DB_URL: "postgresql://{{ archive_db_user }}:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ db_name }}"
         DB_SUPER_URL: "postgresql://postgres:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ db_name }}"
+      delegate_to: "{{ groups.database|first }}"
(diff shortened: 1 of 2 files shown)


copy_database.yml
```diff
@@ -82,7 +82,7 @@
       environment:
         DB_URL: "postgresql://{{ archive_db_user }}:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ db_name }}"
         DB_SUPER_URL: "postgresql://postgres:{{ archive_db_password }}@{{ archive_db_host }}:{{ archive_db_port }}/{{ db_name }}"
-      command: "/var/cnx/venvs/publishing/bin/cnx-db venv"
+      command: "/var/cnx/venvs/archive/bin/cnx-db venv"
```
karenc: Hm... is this meant to run locally on the database server?  Do we need to add `delegate_to: "{{ groups.database|first }}"`?

reedstrm: The underlying code will refuse to install the path-to-venv if it's not a local DB connection, ISTR.

reedstrm: BTW, this came about when Scott ran the db copy to staging. Production _has_ the proper path to the archive venv. Not sure how a path to a non-existent venv got installed.

karenc: Then we need the `delegate_to` line

PR #493, opened 2018-02-05
contentXX servers should be production-like

reedstrm:
Confirmed with @openstaxalina, the contentXX environments/servers should be production servers: they will be used by XML vendors for content work.

AFAIK, content01 and content02 have been deployed, but not yet released to vendor hands. Now is the time to fix them. :-)

### DIFF
--- a/environments/content01/files
+++ b/environments/content01/files
@@ -1,1 +1,1 @@
-../__dev_envs/files
+../__prod_envs/files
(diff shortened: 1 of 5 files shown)


dtwilliamson:
I thought content01 was already in use. I'll confirm that.

dtwilliamson:
It had been used by Katalyst, but is now idle.

PR #500, opened 2018-02-13
rework google-site-verification files

reedstrm:
- handle more than verification filename in vault
  - generate file from name, rather than copying, so no requirement to copy file to bastion[2]


reedstrm:
Woohoo! figured out how to test this. Actually works, now.

reedstrm:
This is needed to avoid regressing the google-site-verification (for https) next time we push webview.

roles/webview/tasks/google_site_verification.yml
```diff
@@ -1,28 +1,17 @@
...
   when:
-    - "google_site_verification_filename is defined"
-    - "google_site_verification_stat.stat.exists"
+    - "vault_google_site_verification_filenames is defined"
   become: yes
   copy:
-    src: "{{ inventory_dir }}/secrets/google/{{ google_site_verification_filename }}"
-    dest: "/var/www/webview/{{ google_site_verification_filename }}"
+    content: "google-site-verification: {{ item }}"
+    dest: "/var/www/webview/{{ item }}"
```
dtwilliamson: this doesn't need to be `.html`?

reedstrm: It has it - the vaulted value is the full filename - which is needed in both places.

PR #502, opened 2018-02-15
Robots.txt refactor 

reedstrm:
move `robots_txt` defs to prod environment, and provide "don't crawl me, bro" defaults for all other environments


reedstrm:
This is part of the Google cnx.org critsit work: this will make sure that future deploys to non-production environments are blocked from search indexing

PR #509, opened 2018-02-26
adding tempCleaner to cnx-deploy pdf_gen role

scb6:


roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+import re
+
+base = r"/tmp"
+now = time.time()
+
+for dirpath, dirnames, files in os.walk(base):
+    for dirname in dirnames:
+        if "tmp" in dirname.lower():
+            f = os.path.join(dirpath, dirname)
+            if os.stat(f).st_mtime < now - 7*86400:
```
reedstrm: Let's not have magic numbers. This should be a config set at the top, or passed in as a param

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+import time
+import shutil
+import re
+
+base = r"/tmp"
+now = time.time()
+
+for dirpath, dirnames, files in os.walk(base):
+    for dirname in dirnames:
+        if "tmp" in dirname.lower():
```
reedstrm: Hmm, pretty sure this matches `/tmp/my_tmp_thing`, which differs from what you spec'ed above (matching only `/tmp/tmp*`) Could be rescued with `if dirname.lower().startswith("tmp"):` (or change the spec)

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+                shutil.rmtree(f)
+                # print "### Delete dir: " + f
+
+for g in os.listdir(base):
+    g = os.path.join(base, g)
+    r = re.compile('modexport|col')
+    s = re.compile('stderr|stdout')
+    if r.search(g):
+        if s.search(g):
+            if os.stat(g).st_mtime < now - 14*86400:
```
reedstrm: more magic numbers

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+                os.remove(g)
+                # print "### Delete: " + g
```
reedstrm: This whole stanza feels not very DRY (Don't Repeat Yourself)
I'm thinking this should be structured as a function you call w/ 3 params: a base, a regex, and an age. You get back an iterator then operate in that. Config is at the top w/ a dictionary keyed on the regex, values are the age. __main__ function that loads config, loops on key,value pairs and deletes.

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+    for dirname in dirnames:
+        if "tmp" in dirname.lower():
+            f = os.path.join(dirpath, dirname)
+            if os.stat(f).st_mtime < now - 7*86400:
+                shutil.rmtree(f)
+                # print "### Delete dir: " + f
+
+for g in os.listdir(base):
+    g = os.path.join(base, g)
+    r = re.compile('modexport|col')
```
reedstrm: recompiling fixed regexs inside a loop. And two of them? Move this out above the loop, and make it a single regex: `r=re.compile('(modexport|col).*\.std(err|out)')`
match it against os.listdir(base) directly, then only do the `os.path.join` for the one's we're going to delete.

roles/pdf_gen/templates/opt/tempCleaner.py
reedstrm: On third read, I see that we're spec'ed a little tighter, this is handling two cases: tempdirs, and directly-in-base files. I guess this could stay as two stanzas, in that case. But do move the magic numbers into constants at the top

reedstrm:
Oh one last thing - I see there's a `systemd-tmpfiles` gizmo that does systemwide tempfile aging via some service. Should we be using that?

roles/pdf_gen/templates/opt/tempCleaner.py
dtwilliamson: or use Python's `datetime` / `timedelta`

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+
+base = r"/tmp"
+now = time.time()
+
+for dirpath, dirnames, files in os.walk(base):
+    for dirname in dirnames:
+        if "tmp" in dirname.lower():
+            f = os.path.join(dirpath, dirname)
+            if os.stat(f).st_mtime < now - 7*86400:
+                shutil.rmtree(f)
```
dtwilliamson: no exception handling for permission denied (also needed for other operations).

roles/pdf_gen/templates/etc/cron.daily/tempCleaner
```diff
@@ -0,0 +1 @@
+sudo -u www-data /opt/tempCleaner.py
```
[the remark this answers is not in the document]
dtwilliamson: cron.daily jobs are run by root - no need for sudo.

roles/pdf_gen/templates/opt/tempCleaner.py
```diff
@@ -0,0 +1,45 @@
...
+
+import os
+import time
+import shutil
+import re
+
+base = r"/tmp"
+now = time.time()
+
+for dirpath, dirnames, files in os.walk(base):
```
dtwilliamson: `walk()` is recursive. All we really care about, I think, is the directories that start with `tmp` that are at the top of `/tmp`.

roles/pdf_gen/templates/opt/tempCleaner.py
dtwilliamson: This should also check to see if it's a file: `os.path.isfile()`.

roles/pdf_gen/templates/opt/tempCleaner.py
dtwilliamson: but if this is only run by root, then less of a problem.

roles/pdf_gen/tasks/main.yml
```diff
@@ -17,3 +17,18 @@
   supervisorctl:
     name: pdf_gen
     state: restarted
+
+- name: Create tempCleaner.py script
+  template: 
+    src: opt/tempCleaner.py
+    dest: /opt/tempCleaner.py
```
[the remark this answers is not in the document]
reedstrm: yeah, I think this is just a temp place to play with it. At least, I hope it is. :_)

roles/pdf_gen/templates/opt/tempCleaner.py
reedstrm: Oh if we only care about the top level, we _can_ collapse the two cases.

PR #511, opened 2018-02-28
Removing the  zabbix role from deployment scripts

scb6:
Because the Zabbix configuration will be managed by a separate set of scripts, we do not want the deployment scripts over writing them, thus we are removing them. (I did not see where they were called for cnx deploy, so either I missed something in my grepping or they were not called)


reedstrm:
this PR is a bit odd: "Files changed: 0"
It's three merge commits?

reedstrm:
I'm interested in the plan for the monitoring management. It seems to me to make sense to include the service-specific management config in each service deployment, if at all possible. This avoids deploying unmonitored services. However, I can see that the technical details of the particular monitoring system may make this hard.

reedstrm:
Umm, I think I understand now. This PR adds, then removes, the zabbix role/tempate stuff. There is no net change, so either it's automatically approved, or there's nothing to approve.

reedstrm:
closing since there's no impact. Don't need these commits on master.

PR #519, opened 2018-03-02
Modify python3 role to use python 3.6

pumazi:
The wrong python version is installed. Python >=3.6 is required and Python 3.5 was installed. This change corrects that.

We'll need to `rm -rf /var/cnx/venvs/press` before a deploy of this.

This is an untested commit.


reedstrm:
Hmm, is there any way to make this 3.5 friendly?

PR #520, opened 2018-03-05
DRY out webview version pin

reedstrm:
And bump version to v0.25.1


reedstrm:
Since I had to do this twice this release, make it a single file. Next step is convincing jenkins to auto-generate the PR

PR #521, opened 2018-03-05
Archive robots txt

reedstrm:
This adds a path for serving a different `robots.txt` for `archive.cnx.org`, and templating for it, under control of cnx-deploy.

### DIFF
--- a/roles/robots_txt/tasks/main.yml
+++ b/roles/robots_txt/tasks/main.yml
@@ -1,11 +1,20 @@
 ---
 # Add cnx robots.txt
 
-- name: Add robots.txt
+- name: Add top level robots.txt
   become: yes
   template:
     src: var/www/webview/robots.txt
     dest: /var/www/webview/robots.txt
+    owner: www-data
+  tags:
+    - robots_txt
+
+- name: Add archive robots.txt
+  become: yes
+  template:
+    src: var/www/robots.txt
+    dest: /var/www/robots.txt
     owner: www-data
   tags:
     - robots_txt
--- a/roles/robots_txt/templates/var/www/robots.txt
+++ b/roles/robots_txt/templates/var/www/robots.txt
@@ -1,0 +1,20 @@
+{% if archive_robots_txt.sitemap_name is defined %}
+Sitemap: https://{{ frontend_domain }}/{{ robots_txt.sitemap_name }}
+{% endif %}
+{% for user_agent in archive_robots_txt.user_agents %}
+User-agent: {{ user_agent.name }}
+{% if user_agent.crawl_delay|default(None) %}
+Crawl-delay: {{ user_agent.crawl_delay }}
+{% endif %}
+{% if archive_robots_txt.disallow_list is defined %}
+{% for disallow in archive_robots_txt.disallow_list %}
+Disallow: {{ disallow }}
+{% endfor %}
+{% endif %}
+{% if archive_robots_txt.allow_list is defined %}
+{% for allow in archive_robots_txt.allow_list %}
+Allow: {{ allow }}
+{% endfor %}
+{% endif %}
+
+{% endfor %}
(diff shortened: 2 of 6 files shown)


roles/robots_txt/tasks/main.yml
```diff
@@ -1,11 +1,20 @@
 ---
 # Add cnx robots.txt
 
-- name: Add robots.txt
+- name: Add top level robots.txt
```
dtwilliamson: "top level" and it's a directory below the dir which holds webview's robots.txt. Not a biggie but maybe "Add webview robots.txt"?

roles/robots_txt/templates/var/www/robots.txt
```diff
@@ -0,0 +1,20 @@
...
+Crawl-delay: {{ user_agent.crawl_delay }}
+{% endif %}
+{% if archive_robots_txt.disallow_list is defined %}
+{% for disallow in archive_robots_txt.disallow_list %}
+Disallow: {{ disallow }}
+{% endfor %}
+{% endif %}
+{% if archive_robots_txt.allow_list is defined %}
+{% for allow in archive_robots_txt.allow_list %}
+Allow: {{ allow }}
```
dtwilliamson: I wonder why the webview robots.txt doesn't have an Allow section.

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -176,15 +176,17 @@ sub vcl_recv {
 
     # cnx rewrite archive  - specials served from nginx statically
     if (req.http.host ~ "^{{ arclishing_domain }}" || req.url ~ "^/sitemap.*.xml") {
-        if ( req.method == "POST" || req.method == "PUT" || req.method == "DELETE" || req.url ~ "^/(publications|callback|a|login|logout|moderations|feeds/moderations.rss|contents/.*/(licensors|roles|permissions))") {
+        
+        if (req.url  == "/robots.txt" || req.url ~ "^/specials") {
```
dtwilliamson: No regex anchor needed for robots?

I seem to remember this block having moved before.

roles/robots_txt/templates/var/www/robots.txt
pumazi: Seems like it might be a good idea to collapse this into a single temple file. But It's ok as is as well.

reedstrm: Yeah, I thought about making it a single template, but my ansible-fu is weak: I didn't know how to change the variable values for different calls to template. Checking stackoverflow, this seems to be a common ansible complaint/weakness.

roles/varnish/templates/etc/varnish/varnish.vcl
reedstrm: I made it `==` since only the one URL is valid

reedstrm: As to the move, this just seemed neater to me. Take care of the simple cases, leave the complex one to later

roles/robots_txt/tasks/main.yml
reedstrm: Actually, they're both top level when served via http- one from cnx.org, one from archive.cnx.org. Confusingly, the cnx.org one ("top level" for domain) is served from lower down the filesystem tree, while the archive.cnx.org one is (here) now being served from the nginx webroot. Sane? Not really. I considered adding an 'archive' subdir, for the one file, but that would have required even more changes (an extra stanza in nginx config as well)

kerwinso:
Related Trello card: https://trello.com/c/uadCIRn4/401-serve-a-different-robotstxt-for-archivecnxorg

roles/robots_txt/tasks/main.yml
dtwilliamson: I was just suggesting clarifying the comment.

roles/robots_txt/templates/var/www/robots.txt
reedstrm: I jus spent 30 min poking at what should have been a sane solution (using include_var) but ansible is not sane.

reedstrm:
This was developed/tested on dev, so I know the routing works.

roles/robots_txt/templates/var/www/robots.txt
dtwilliamson: one of us, one of us

PR #522, opened 2018-03-09
Expose the press instance(s) to the outside world

pumazi:
- This assigns a default port to the Press service, similar to how it's done with the other apps
- This punches a hole in the firewall to allow the request pipeline to communicate with Press instance(s). Currently, the service is blocked off from communication with the outside world.
- This exposes the Press service on cnx.org.

### DIFF
--- a/roles/press/templates/etc/supervisor/conf.d/press.conf
+++ b/roles/press/templates/etc/supervisor/conf.d/press.conf
@@ -4,4 +4,4 @@ directory=/var/cnx/apps/press
 numprocs={{ hostvars[inventory_hostname].press_count|default(1) }}
 process_name=%(program_name)s-%(process_num)s
 environment=SHARED_DIR="/var/cnx/apps/press/var",DB_URL="postgresql://{{ press_db_user }}:{{ press_db_password }}@{{ press_db_host }}:{{ press_db_port }}/{{ press_db_name }}",DB_SUPER_URL="postgresql://postgres@{{ press_db_host }}:{{ press_db_port }}/{{ press_db_name }}",SENTRY_DSN="{{ sentry_dsn|default() }}"
-command=/var/cnx/venvs/press/bin/gunicorn -b 0.0.0.0:6543 --access-logfile - --error-logfile - -n press --reload wsgi:app
+command=/var/cnx/venvs/press/bin/gunicorn -b 0.0.0.0:{{ press_base_port|default(default_press_base_port) }} --access-logfile - --error-logfile - -n press --reload wsgi:app
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -140,6 +140,16 @@ sub vcl_init {
 {% endfor %}
 {% endfor %}
 
+    # Define the Press cluster
+    new press_cluster = directors.hash();
+{% for host in groups.press %}
+{% for i in range(0, hostvars[host].press_count|default(1), 1) %}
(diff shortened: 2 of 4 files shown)


roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -160,6 +170,11 @@ sub vcl_recv {
         return (pass);
     }
 
+    if (req.url ~ "^/api/") {
```
reedstrm: So, press is claiming cnx.org/api ? That's a pretty serious land-grab for something that has been described as temporary. Are we rethinking the likely lifespan of press? Do we intend it to take over everything publishing is currently doing?

Also, so far we've kept cnx.org for humans to interact with, archive.cnx.org as the machine interface. Would that be a better choice for where press's API lives, whatever it's route?

roles/press/templates/etc/supervisor/conf.d/press.conf
```diff
@@ -4,4 +4,4 @@ directory=/var/cnx/apps/press
 numprocs={{ hostvars[inventory_hostname].press_count|default(1) }}
 process_name=%(program_name)s-%(process_num)s
 environment=SHARED_DIR="/var/cnx/apps/press/var",DB_URL="postgresql://{{ press_db_user }}:{{ press_db_password }}@{{ press_db_host }}:{{ press_db_port }}/{{ press_db_name }}",DB_SUPER_URL="postgresql://postgres@{{ press_db_host }}:{{ press_db_port }}/{{ press_db_name }}",SENTRY_DSN="{{ sentry_dsn|default() }}"
-command=/var/cnx/venvs/press/bin/gunicorn -b 0.0.0.0:6543 --access-logfile - --error-logfile - -n press --reload wsgi:app
+command=/var/cnx/venvs/press/bin/gunicorn -b 0.0.0.0:{{ press_base_port|default(default_press_base_port) }} --access-logfile - --error-logfile - -n press --reload wsgi:app
```
reedstrm: Are we experimenting w/ a different server here? What advantages does gunicorn give us over waitress? I've seen comments re: gunicorn allowing massive parallel hammering of the DB server (auto-scaling of number of workers) Do we need to control for that? Can't we specify gunicorn in the ini file, and still launch via pserve? (for a consistent devops experience :-)

roles/varnish/templates/etc/varnish/varnish.vcl
pumazi: It's the lowest common match item at the moment. Things can always change.

We can put it on archive.cnx.org if that feels better for you. I'd like to see everything under one domain, but it's apples and oranges. 

I'm personally seeing press as something more long-term now. But this is a discussion to have outside the scope of this PR.

roles/press/templates/etc/supervisor/conf.d/press.conf
pumazi: This is outside the scope of this PR, but feel free to create an issue for it.

pumazi:
You are correct. There isn't anything here to make this specific to the 'dev' env. It might help to put a condition around the vanish rule. I'll go ahead and do that.

roles/press/templates/etc/supervisor/conf.d/press.conf
reedstrm: Ah I see that it's already gunicorn. Definitely want to think about that.

reedstrm:
Actually, don't bother - the next step after testing on dev will be a request to test on a contentXX server, which is production style as well. If we don't make it a code change to deploy to a diff environment, then devops can handle it.

PR #557, opened 2018-04-12
switch on stub accounts for all environments

reedstrm:
There is currently no need for publishing to talk to accounts. Turn on stub accounts instead.


scb6:
The code looks OK to me, but I do not know much about what these change will do to the code base. (i.e. Mulich's comments)

reedstrm:
Agreed - I did the minimum to deploy, but spread it across all the environments. I think perhaps the right fix for _today_ (code freeze) may be to strip this off the production environs - @pumazi is correct, we should vault the credentials. Who knows how to do that, @scb6? @DarkPrinceFrost ?

pumazi:
Oi, lots of environments... I'm going to create a ./group_vars/vault.yml to hit all the environments. Let me know if that's an issue for you guys.

scb6:
Looks good to me.

PR #584, opened 2018-05-03
fix incorrect module

dtwilliamson:


dtwilliamson:
@reedstrm I believe this is ready to merge.

PR #599, opened 2018-05-17
Adding updates to iptable role, including opening a port for zeo

scb6:


iptables.yml
```diff
@@ -5,7 +5,7 @@
   roles:
     - role: iptables
       when:
-        - install_iptables is defined
+        - install_iptables|bool == true
```
pumazi: I'm guessing this is because command line options are not cast to native values the same way they would be in yaml? Just curious about this... :-)

scb6: Your guess is right, however, I wrote it this way based on an old post that, while "right" is not best way to do this any more.  I will need to make an update before merge. Thanks for calling attention to it.

PR #604, opened 2018-05-24
Change the accounts_stub_user variable to match

scb6:
Change the accounts_stub_user variable to match in the vault file to match what is being called in the group_vars/all.yml.


scb6:
Hmm.  This change still fails with the same error: 
```
TASK [publishing_common : render configuration] **********************************************
failed: [dev00.cnx.org] (item=etc/cnx/publishing/app.ini) => {"changed": false, "item": "etc/cnx/publishing/app.ini", "msg": "AnsibleUndefinedVariable: 'vault_accounts_stub_users' is undefined"}
```

pumazi:
Yep, that seems odd, but it doesn't surprise me, because ansible. I guess vault files in the global group_vars space isn't read? Or rather, there is no group called 'vault', so it's not applying the variables. IIRC, `.../group_vars/{group-name}(.yml | /{any-filename}.yml)`

kerwinso:
Related to this (I think): https://github.com/Connexions/cnx-deploy/pull/557

Dantemss:
Remove this line: https://github.com/Connexions/cnx-deploy/blob/0d5879c5054cc9029088aa6ad4e82701d72ec21d/group_vars/all.yml#L76

Dantemss:
The variable in the vault might have a typo

PR #617, opened 2018-05-29
Added zeopack wrapper script to automatically delete the old Data file when packing is successful

Dantemss:
Tested successfully on -dev (the script was manually executed, the cron job was not tested but it should just work :tm:).

0755 means world readable, but that's what the cnx-buildout dir is currently set to.

### DIFF
--- a/roles/zeopack/templates/zeopack-wrapper.sh
+++ b/roles/zeopack/templates/zeopack-wrapper.sh
@@ -2,6 +2,6 @@
 {{ zeopack_filepath }}
 exit_status=$?
 if [ $exit_status -eq 0 ]; then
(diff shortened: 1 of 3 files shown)


roles/zeopack/templates/zeopack-wrapper.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+{{ zeopack_filepath }}
+exit_status=$?
+if [ $exit_status -eq 0 ]; then
```
reedstrm: The explicit test against 0 is not necessary, but is clear. I'll allow it. :-)

PR #621, opened 2018-05-31
Added play and updated play to include press in history.txt

scb6:


update_versions.yml
```diff
@@ -65,6 +65,16 @@
       with_items:
         - "{{ groups.zope|first }}"
 
+- name: get press version
+  hosts: frontend
+  tasks:
+    - shell: grep '^[0-9]' /var/cnx/apps/press/docs/source/changes.rst  | head -n 1
+      register: press_version
```
pumazi: Is there any chance this could conflict with the `press_version` variable?

kerwinso: Apparently so, see #636.

PR #624, opened 2018-06-01
Reverting press n varnish

pumazi:
- Remove press from the prod and staging environments
- Revert the varnish "band-aid"

### DIFF
--- a/environments/prod/inventory
+++ b/environments/prod/inventory
@@ -94,7 +94,6 @@ prod06
 backup2
 
 [press:children]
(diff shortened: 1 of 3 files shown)


kerwinso:
@pumazi A few questions:
1. What about content01 through content05? See https://github.com/Connexions/cnx-deploy/pull/589/files.
1. Do all the `vars.yml` files need reverting as well? See PR linked above.
1. Will Sentry integration be affected?

PR #636, opened 2018-06-08
Change press Version variable name

scb6:


pumazi:
This adjustment was made because we were experiencing something odd with the `press_version` in the following code. For some unknown reason the version looked to be a registered variable of ansible task output. The only naming collision was what this PR changes.

https://github.com/Connexions/cnx-deploy/blob/99af1902aaa8765fff7d1d9732b8f3f1a62778e3/roles/press/tasks/main.yml#L80-L88

PR #646, opened 2018-06-15
restore some legacy routes to varnish config - needed for latex pdf g…

reedstrm:
…eneration

Seems we _still_ can strip all the legacy out of the non-zope varnish: PDFgen expects certain things there.


reedstrm:
The only one of these I know for _sure_ that is needed is the RDF route, but I added back the other three just in case there are other scripts that assume they exist.

PR #652, opened 2018-06-20
set some timeouts for NFS mounts so archive doesn't hang if exports g…

reedstrm:
…o away, like in today's critsit.


reedstrm:
the options doc is in `man nfs` - not sure how to link that. :-/

PR #663, opened 2018-07-03
Remove Press from the next prod-like deployment

pumazi:
A problem in the legacy software is preventing us from releasing. The problem stems from reading data from the database rather than legacy's usual role of writing to it.

### DIFF
--- a/environments/content01/inventory
+++ b/environments/content01/inventory
@@ -52,7 +52,7 @@ content01
 [backup:children]
 
 [press:children]
-content01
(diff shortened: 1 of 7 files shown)


kerwinso:
Please HOLD on merging this until CNX team approves, thank you.

kerwinso:
7/5/18 Decision: Release with press _in_. Update Neb to prevent publishing to prod. Tell CMs they cannot publish to prod.

PR #669, opened 2018-07-10
added logic to open 2049 for certain groups.

scb6:


roles/iptables/templates/etc/iptables
```diff
@@ -29,6 +29,9 @@
 {% for ipaddr in groups.legacy_frontend -%}
 -A INPUT -p tcp -m tcp -m multiport --dports 8080,8888,8991 -m comment --comment "allowed TCP ports for Legacy" -j ACCEPT
 {% endfor -%}
+{% for ipaddr in [groups.legacy_frontend, groups.frontend, groups.archive, groups.pdf_gen] -%}
```
reedstrm: Actually, I'm pretty sure all we need is `groups.nfs_connected` In fact, that is a super-set of the above groups, with in addition a couple I missed.

reedstrm: Sorry I misled you initially :/

reedstrm:
Ohh, and @DarkPrinceFrost  here's a handy way (using exactly this case as the example!) to have different stanzas for the different servers:
https://stackoverflow.com/questions/31347921/how-to-structure-a-template-file-which-might-be-different-for-groups-of-hosts-in

dtwilliamson:
How is that different other than using `if` instead of `for`?

reedstrm:
@dtwilliamson  - it'd be "if" one group then "for" over another: in this case, if nfs_server, loop over nfs_clients

PR #674, opened 2018-07-11
Add openstaxcms to settings.js

Dantemss:
I guessed the desired CMS urls based on the other urls in each config.
NOTE: several URLs in the `des` environment do not exist

Goes with https://github.com/Connexions/webview/pull/1713


dtwilliamson:
I don't think the des environment is used any longer.

reedstrm:
BTW, for all the other `-qa` settings in those files, we need to talk to @openstaxalina first, before changing anything.

Dantemss:
@reedstrm content, devb and katalyst envs changed to prod exercises and cms

reedstrm:
Ignoring the pyup security complaint - this PR doe not bring in that dependecy.

PR #691, opened 2018-07-23
add frozen list of press requirements for prod-like envs

reedstrm:
Conditionally install the frozen list, if it exists, in favor of the in-repo lists. Also, always update dependencies, if possible, for dev environments


reedstrm:
The pyup safety complaint about PyYAML is not resolvable, yet - the developers pulled the 4.1 release, so there is no safe version to upgrade to.

reedstrm:
Finally got all the ansible conditionals working the way it should. :-) Thanks for the 'press.yml' playbook, made iterating on `dev` and `vendor-staging` a lot faster.

PR #694, opened 2018-07-24
add press full versions to python-versions.txt

reedstrm:
Had this lying around - @kerwinso would like it.


reedstrm:
has been deployed to dev - check https://dev.cnx.org/python-version.txt

PR #705, opened 2018-08-01
Add Press worker processes

pumazi:
This adds configuration for Celery worker processes that work for the Press service. The settings for this work are in #704

The first portion of this is in #703 . Feel free to close that if you can read this one given the moves and all.


roles/press_common/tasks/main.yml
```diff
@@ -0,0 +1,168 @@
...
+# +++
+# Prerequisites
+# +++
+
+- name: install system level dependencies
+  become: yes
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
```
dtwilliamson: build tools? can't whatever be built once elsewhere and a resulting package installed instead?

roles/press_common/tasks/main.yml
```diff
@@ -0,0 +1,168 @@
...
+  file:
+    path: "/var/cnx/apps"
+    state: directory
+    recurse: yes
+    owner: www-data
+
+- name: checkout codebase
+  become: yes
+  become_user: www-data
+  git:
```
dtwilliamson: See previous comment

roles/press_common/tasks/main.yml
pumazi: This is existing logic that has been moved. I'm not opposed to discussing that at another time though. 😸

roles/press_worker/templates/etc/supervisor/conf.d/press_celery_workers.conf
```diff
@@ -0,0 +1,12 @@
+{% for i in range(0, hostvars[inventory_hostname].press_worker_count|default(1), 1) %}
```
Dantemss: Isn't a default of 1 a little small? Especially if the tasks can take a really long time.

pumazi: None of the currently implemented task will take a long time. Shouldn't take but 1-2sec max.

PR #713, opened 2018-08-03
Assign Press worker hosts

pumazi:
I unfortunately forgot to define `press_worker` in the inventory files and assign hosts to that host group. This rectifies that mistake.

### DIFF
--- a/environments/content02/inventory
+++ b/environments/content02/inventory
@@ -54,32 +54,36 @@ content02
 [press:children]
 content02
 
+[press_worker:children]
+content02
+
 # Groups of groups
 
 [broker_connected:children]
 publishing
 channel_processing
 publishing_worker
+press
+press_worker
 
 [db_connected:children]
 archive
 publishing
 channel_processing
 publishing_worker
-authoring
 zclient
 pdf_gen
 backup
 press
+press_worker
 
 [nfs_connected:children]
-archive
-publishing
 legacy_frontend
 frontend
 zclient
 pdf_gen
 press
+press_worker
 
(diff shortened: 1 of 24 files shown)


pumazi:
@Dantemss, that's the same thing I questioned in #cnx-deploy chat the other day. 😸 

And yep, no tests...

environments/content02/inventory
```diff
@@ -54,32 +54,36 @@ content02
...
 publishing_worker
+press
+press_worker
 
 [db_connected:children]
 archive
 publishing
 channel_processing
 publishing_worker
-authoring
```
reedstrm: This is fine - authoring needs to be stripped out

environments/content02/inventory
```diff
@@ -54,32 +54,36 @@ content02
...
 zclient
 pdf_gen
 backup
 press
+press_worker
 
 [nfs_connected:children]
-archive
```
reedstrm: This is not: we still need access to the NFS mounted file exports - archive uses FS access to list what downloadable files are available

environments/content03/group_vars/all/vars.yml
```diff
@@ -16,8 +16,8 @@ press_db_user: "{{ archive_db_user }}"
 press_db_password: "{{ archive_db_password }}"
 press_db_host: "{{ archive_db_host }}"
 press_db_port: "{{ archive_db_port }}"
-press_broker_user: "{{ vault_press_broker_user }}"
```
reedstrm: This too should probably stay vaulted

PR #725, opened 2018-08-06
restore nfs children for archive and publishing

reedstrm:


pumazi:
It might have been good to add press to this list as well, but o well.

PR #728, opened 2018-08-07
Initial HAProxy and nginx config optimizations that don't require software updates

Dantemss:
HAProxy's `option httpclose` is deprecated, the replacement being `option forceclose`, which would prevent keep-alive connections. However, `option http-server-close` can be used instead to obtain a keep-alive connection towards the client, while shielding the application servers from them.

Also removed some dead code (the `xxx_hardcoded_zclients` variable seems to no longer be set anywhere).

nginx config changes:

- Removed files/etc/nginx/nginx.conf (was unused)
- Copied the nginx config that was currently in the servers (prod) into the nginx role (but uncommented-out the gzip config)
- Added task to copy this nginx config to the servers during deployment
- Tested gzip compression on dev.cnx.org (main.js goes from 1.3MB to 300kB).


Dantemss:
Interesting... the nginx.conf file is actually never written to the server, it looks like. The servers already had tcp_nopush and tcp_nodelay on. But some gzip options are commented out there.

reedstrm:
Do you think we should fully-template the nginx config? I think it currently edits-whatever-is-there from the default nginx install.

Dantemss:
@reedstrm @pumazi made the following changes:

- Removed files/etc/nginx/nginx.conf (was unused)
- Copied the nginx config that was currently in the servers (prod) into the nginx role (but uncommented-out the gzip config)
- Added task to copy this nginx config to the servers during deployment

The purpose of this change (besides allowing us to make changes to the config via this repo) is to make sure all the servers have the same config.

PR #738, opened 2018-08-13
Set the Cache-Control headers for webview and static files in nginx

Dantemss:
Since nginx is the closest thing we have to a backend for webview and static files.

Caching is hard so nothing is set in stone. Comments welcome. Probably good to do a test deployment to -dev before merging.

### DIFF
--- a/roles/static_files/templates/etc/nginx/sites-available/static_files
+++ b/roles/static_files/templates/etc/nginx/sites-available/static_files
@@ -16,7 +16,7 @@ server {
     }
 
     location ~ ^/files/(.+)$ {
-        expires   1y;
+        expires   1m;
         autoindex off;
         set $fname $1;
         add_header Content-Disposition "attachment; filename=$fname";
--- a/roles/static_files/templates/etc/nginx/sites-available/static_files
+++ b/roles/static_files/templates/etc/nginx/sites-available/static_files
@@ -8,12 +8,15 @@ server {
     listen   8080;
     listen   [::]:8080 default ipv6only=on;  ## listen for ipv6
     port_in_redirect off;
+    expires          -1;
 
     location /specials {
+        expires    1m;
         add_header Access-Control-Allow-Origin $cors_header;
     }
 
     location ~ ^/files/(.+)$ {
+        expires   1y;
         autoindex off;
         set $fname $1;
         add_header Content-Disposition "attachment; filename=$fname";
--- a/roles/webview/templates/etc/nginx/sites-available/webview
+++ b/roles/webview/templates/etc/nginx/sites-available/webview
@@ -5,6 +5,7 @@ server {
     root            /var/www/webview;
     index           index.html;
     try_files       $uri $uri/ /index.html;
+    expires         -1;
 
     location /ping {
         add_header Content-Length 0;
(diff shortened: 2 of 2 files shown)


kerwinso:
Dante says: delay this till we can test better.

Dantemss:
I think one of the entries here has to be changed and made way shorter (either /files or /exports, don't remember which). Will ask @reedstrm tomorrow.

roles/static_files/templates/etc/nginx/sites-available/static_files
```diff
@@ -8,12 +8,15 @@ server {
     listen   8080;
     listen   [::]:8080 default ipv6only=on;  ## listen for ipv6
     port_in_redirect off;
+    expires          -1;
 
     location /specials {
+        expires    1m;
```
reedstrm: specials change so rarely, we can make this much longer - Though expires only will control time till recheck. Setting _something_ is better than nothing.

roles/static_files/templates/etc/nginx/sites-available/static_files
```diff
@@ -8,12 +8,15 @@ server {
     listen   8080;
     listen   [::]:8080 default ipv6only=on;  ## listen for ipv6
     port_in_redirect off;
+    expires          -1;
```
reedstrm: ?? Pre-expired, or infinite?

roles/static_files/templates/etc/nginx/sites-available/static_files
```diff
@@ -8,12 +8,15 @@ server {
...
 
     location /specials {
+        expires    1m;
         add_header Access-Control-Allow-Origin $cors_header;
     }
 
     location ~ ^/files/(.+)$ {
+        expires   1y;
```
reedstrm: this would be great, except for the dang PDFs that get replaced with the same filename when a new one is uploaded. Again, expires controls recheck, so we could live w/ shorter, as long as we implement ETAGs  as well.

roles/webview/templates/etc/nginx/sites-available/webview
```diff
@@ -68,6 +69,7 @@ server {
     # and if they do not exist, fall back to cnx-authoring
     # from http://linuxplayer.org/2013/06/nginx-try-files-on-multiple-named-location-or-server
     location ~ ^/resources/.+ {
+        expires          1y;
```
reedstrm: url _is_ a content hash, so 1year is just fine. :-)

roles/static_files/templates/etc/nginx/sites-available/static_files
Dantemss: Default is no cache, unless explicitly set elsewhere (which they should all be).

roles/static_files/templates/etc/nginx/sites-available/static_files
Dantemss: Per our discussion, after 1 minute the browser will simply recheck the etags and the request is answered by nginx, so it's fast enough that we should be able to keep this at 1 minute.

Dantemss:
Changed the PDF from 1 year to 1 minute (probably didn't matter much anyway, since downloads are usually not cached, but good to be safe).

roles/static_files/templates/etc/nginx/sites-available/static_files
reedstrm: I guess that answers my question: -1 is pre-expired, so won't be cached.

PR #740, opened 2018-08-16
Cleanup and tweak the varnish cache config

Dantemss:
This goes with the other PRs that push the Cache-Control stuff to webview/archive nginx/waitress. If you see anything that looks weird, please point it out, since this is a pretty big change.

### DIFF
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -212,13 +212,11 @@ sub vcl_recv {
     # cnx rewrite stub login form
     elsif (req.url ~ "^/stub-login-form") {
         set req.backend_hint = publishing_cluster.backend(req.http.cookie);
-        return (pass);
     }
 {% endif %}
     # cnx rewrite archive
     elsif (req.url ~ "^/a/") {
         set req.backend_hint = publishing_cluster.backend(req.http.cookie);
-        return (pass);
     }
     elsif (req.url ~ "^/resources/") {
         # Note: Medium-sized static files served directly from waitress
@@ -239,7 +237,6 @@ sub vcl_recv {
         }
         elsif ( req.method == "POST" || req.method == "PUT" || req.method == "DELETE" || req.url ~ "^/(publications|callback|a|login|logout|moderations|feeds/moderations.rss|contents/.*/(licensors|roles|permissions))") {
             set req.backend_hint = publishing_cluster.backend(req.http.cookie);
-            return (pass);
         }
         else {
             set req.backend_hint = archive_cluster.backend();
--- a/roles/varnish/templates/etc/varnish/varnish.vcl
+++ b/roles/varnish/templates/etc/varnish/varnish.vcl
@@ -3,24 +3,22 @@ vcl 4.0;
 import std;
 import directors;
 
-# This configuration uses inline C, so you must run the program with
-# the include C parameter: -r vcc_allow_inline_c
-
 # This is a basic VCL configuration file for varnish.  See the vcl(7)
 # man page for details on VCL syntax and semantics.
-# C{
-#         #include <string.h>
-#         #include <stdlib.h>
-#         #include <time.h>
 
-#         void TIM_format(double t, char *p);
-#         double TIM_real(void);
-#         time_t TIM_parse(const char *p);
-# }C
-
+# IP addresses blocked from performing any requests
 acl block {
 {% for ip_addr in blocked_ip_addresses|default([]) %}
 "{{ ip_addr }}";
+{% endfor %}
+}
+
+# IP addresses allowed to use the PURGE http verb
+acl purge {
+    "localhost";
+    "127.0.0.1";
+{% for host in groups.varnish_purge_allowed %}
+    "{{ hostvars[host].ansible_default_ipv4.address }}";
 {% endfor %}
 }
 
@@ -45,7 +43,7 @@ backend legacy_frontend {
 }
 {% endif %}
 
-# static files
+# static_files nginx
 backend static_files {
 .host = "127.0.0.1";
 .port = "8080";
@@ -54,6 +52,7 @@ backend static_files {
 .between_bytes_timeout = 10s;
 }
 
+# webview nginx
 backend webview {
 .host = "127.0.0.1";
 .port = "8081";
@@ -66,6 +65,7 @@ probe archive_probe {
     .expected_response = 404;
 }
 
+# archive waitress
 {% set base_port = archive_base_port|default(default_archive_base_port) %}
 {% for host in groups.archive %}
 {% for i in range(0, hostvars[host].archive_count|default(1), 1) %}
@@ -88,6 +88,7 @@ probe publishing_probe {
     .expected_response = 404;
 }
 
+# publishing waitress
 {% set base_port = publishing_base_port|default(default_publishing_base_port) %}
 {% for host in groups.publishing %}
 {% for i in range(0, hostvars[host].publishing_count|default(1), 1) %}
@@ -105,13 +106,12 @@ backend {{ backend_name }} {
 {% endfor %}
 {% endfor %}
 
-{# FIXME within an if condition because it's not production worthy yet #}
-{% if 'press' in groups and groups['press'] %}
 probe press_probe {
     .url = "/ping";
     .expected_response = 200;
 }
 
+# press waitress
 {% set base_port = press_base_port|default(default_press_base_port) %}
 {% for host in groups.press %}
 {% for i in range(0, hostvars[host].press_count|default(1), 1) %}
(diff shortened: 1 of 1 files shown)


roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -176,89 +165,85 @@ sub vcl_init {
...
-    if (req.http.host ~ "^{{ arclishing_domain }}" || req.url ~ "^/sitemap.*.xml") {
-
+    elsif (req.url ~ "^/resources/") {
+        # Note: Medium-sized static files served directly from waitress
+        # We could pipe, but varnish is probably
+        # more suitable for serving these files than waitress
+        set req.backend_hint = resource_cluster.backend();
+    }
+    elsif ( req.url ~ "^/exports/") {
+        # Note: Large static files served directly from waitress
```
reedstrm: Caching large files blows out the RAM cache, eviciting other small and medium files which are more likely to be served again. We should redirect these and serve from nginx, if possible.  The problem there is that archive has code to look up multiple names, and create a filesystem link to the appropriate other named file (legacy colXXX filenames linked as <UUID> names) We'd need logic to attempt to find via nginx, and if getting a 404, _then_ ask archive. Probably doable w/ some sort of restart. Need to break out the VCL state diagram. Not really a blocker for this PR, since this does seem to describe current behavior (I swear we used to do the other - I think inside nginx itself, somehow …)

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -269,31 +254,39 @@ sub vcl_recv {
...
-            return(hash);
         }
         elsif (req.restarts == 0  && req.url ~ "^/content/.*/module_export_template") {
             set req.backend_hint = legacy_frontend;
-            return(hash);
         }
         elsif (req.restarts == 0  && req.url ~ "^/content/(m[0-9]+)/([0-9.]+)/.*format=pdf$") {
             set req.backend_hint = static_files;
             set req.url = regsub(req.url, "^/content/(m[0-9]+)/([0-9.]+)/.*format=pdf", "/files/\1-\2.pdf");
+            return (pipe);
```
reedstrm: Hmm, `pipe` is pretty strongly discouraged in varnish-lore, especially for what are actual HTTP transactions. I recall us having problems w/ not getting the last few kbytes of some files sometimes, solved by removing some pipes.

Oh btw, this restarts code is an example for the earlier comment - it tries to get the module PDFs from nginx, and if that fails, calls zope, which builds the PDF, serves it, _and_ stores it on the filesystem, for the next call.

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -302,301 +295,142 @@ sub vcl_recv {
         elsif (req.url ~ "^/content/((col|m)[0-9]+)") {
             set req.backend_hint = archive_cluster.backend();
         }
-        // all webview
+        # all webview
         elsif (req.url ~ "_escaped_fragment_=" || req.url ~ "^/$" || req.url ~ "^/?.*" || req.url ~ "^/opensearch\.xml" || req.url ~ "^/version\.txt" || req.url ~ "^/search" || req.url ~ "^/contents$" || req.url ~ "^/(contents|data|exports|styles|fonts|bower_components|node_modules|images|scripts)/" || req.url ~ "^/(about|about-us|people|contents|donate|tos|browse)" || req.url ~ "^/(login|logout|workspace|callback|users|publish|robots.txt)") {
```
reedstrm: at one point this was a complete "whitelist" of URLs we wanted to serve from webview, allowing a fallback below to legacy. We took that fallback out, and not sure we've been maintaining this list (pretty d*mn sure we have not, actually)

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -302,301 +295,142 @@ sub vcl_recv {
...
 sub vcl_hash {
-    hash_data(req.url);
-    if (req.http.host) {
-        hash_data(req.http.host);
-    } else {
-        hash_data(server.ip);
-    }
-    if (req.http.Accept ~ "application/xhtml\+xml" && req.url ~ "^/contents/") {
+    # The response to this archive route varies depending on the Accept header
+    # This works like a normalized Vary: Accept
```
reedstrm: Right, the previous hash of host was because this varnish would receive reqs for 'cnx.org', 'legacy.cnx.org' and even 'archive.cnx.org' - I _think_ we've got that sorted in the higher-order routing now - Hmm, don't we still get 'cnx.org' and 'archive' urls here? We serve very different results … i.e. cnx.org/search?q=string returns the standard webview app page, archive.cnx.org/search?q=string returns the json search results - pretty sure this will return the webview html for a search, as written. (also true for /contents, etc.)

Dantemss:
The builtin `vcl_hash` will hash the host (or ip, if no host).

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -176,89 +165,85 @@ sub vcl_init {
...
 sub vcl_recv {
-
+    # Block any clients with an ip in the block acl
     if (client.ip ~ block) {
         return (synth(403, "Access denied"));
     }
 
+    # Remove multiple trailing /
+    if (req.url ~ "/{2,}$") {
+        return (synth(301, bereq.http.host + regsub(bereq.url, "/{2,}$", "/")));
```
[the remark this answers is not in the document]
Dantemss: I think the idea was to not have the user see the extra `/`. Could do it in javascript, which is able to rewrite the url without reloading the page, but of course that wouldn't happen in varnish.

roles/varnish/templates/etc/varnish/varnish.vcl
```diff
@@ -176,89 +165,85 @@ sub vcl_init {
...
+        }
+        elsif (req.method == "PURGE_REGEXP") {
+          ban("req.url ~ " + req.url);
+          return (synth(200, "Regexp ban added"));
+        }
+    }
+
+    # Routing
     if (req.url ~ "^/ping") {
         set req.backend_hint = webview;
```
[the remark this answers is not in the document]
Dantemss: I added this to the webview nginx config here: https://github.com/Connexions/cnx-deploy/pull/738/files#diff-302b3746a425f79ae7657d334ceb5cc9R8

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: Yes, as you mentioned, I think this is the same as the current behavior.

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: I think you are right. Though piping to nginx would be ideal, the restart may fail if we pipe here. I'll take it out.

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: Should this just become an `else`?

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: Covered by the builtin vcl_hash (it hashes the host or the ip, if no host).

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: Actually that might be dangerous to change. Current behavior seems to be that anything that does not match this whitelist falls through to legacy_frontend (the first backend in the list).

roles/varnish/templates/etc/varnish/varnish.vcl
Dantemss: Actually... even though this code pretends to handle restarts... the only restart that would ever happen would be on `/files/`, since that's the only route that can `return (retry)`

roles/varnish/templates/etc/varnish/varnish.vcl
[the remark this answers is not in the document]
reedstrm: fair enough, leave it in, I guess. :-)

roles/varnish/templates/etc/varnish/varnish.vcl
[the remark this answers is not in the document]
reedstrm: Ah layers. Good.

roles/varnish/templates/etc/varnish/varnish.vcl
reedstrm: yeah, yeah, I guess. Just wanted to write it out. :-)

roles/varnish/templates/etc/varnish/varnish.vcl
reedstrm: ah, good enough.

roles/varnish/templates/etc/varnish/varnish.vcl
reedstrm: if we've missed anything, I suppose QA will let us know. :-)

PR #755, opened 2018-08-23
Downgrade cnxdb 2.3.2 to 2.3.0

reedstrm:
This downgrades cnx-db to remove the abstract html trigger. I've since found that it fires at the wrong time, so instead need to track the module_files_add trigger and find out why it's not doing the expected abstract transforms.


reedstrm:
I'm going to close this - not needed for the existing release, will do it later as part of the proper HTML abstract transformation fix.

PR #815, opened 2018-09-25
Added verification file for new Google Search Properties

scb6:
- http://www.cnx.org
- https://www.cnx.org
- http://archive.cnx.org
- https://archive.cnx.org
- http://legacy.cnx.org
- https://legacy.cnx.org

Question for @reedstrm or whoever else is in the know: Does the verification file for archive and legacy go in the same place and cnx.org?


scb6:
I verified by creating a CNAME record for DNS. This PR is not needed.

PR #908, opened 2018-11-19
moved the press probe to /api/ping

reedstrm:
ibid


reedstrm:
withdrawing this  - I think I'll have  both routes live.

PR #916, opened 2018-11-26
migration needs db write access

reedstrm:
as the title says, probably need write access - not all archive hosts have that. :-)


reedstrm:
@scb6 so, who merges?

PR #920, opened 2018-11-28
revert cnx-press to v7.0.1

reedstrm:
~not ready to deploy this yet~ I don't remember typing that. This _is_ ready to merge and deploy to staging, in my opinion.


reedstrm:
I think this is in fact ready to deploy to staging. After merge it will need to be tagged v2.32.1

PR #985, opened 2019-01-23
Lint check all changes using Travis-CI

pumazi:
This adds `ansible-lint` for linting this project. It fails horrendously right now, but it's a conversation starter none the less. 🥂

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+dist: trusty
+sudo: required
+language: python
+python:
+  - "2.7"
+before_install:
+  - pip install -r requirements.txt
+  - pip install ansible-lint
+script:
+  - ansible-lint main.yml


scb6:
I am on board with adding a linter to start the conversation about getting `cnx-deploy` updated. We can prioritize these issues as they come. I am a little concerned this will create a lot on noise while issues are being created, but as long as this doesn't break the deployment while things are being fixed, I approve.

pumazi:
Sounds like we'll need to get the initial linting completed before we can push this in then. That's tedious task that we'll have to find time to get done.

m1yag1:
Change this into using a pre-commit hook.

Bkblodget:
as a developer i would like to use pre-comit to lint my newly staged changes for cnx-deploy

PR #1102, opened 2019-05-24
Creates static configuration for redirects from CNX to REX

pumazi:
See openstax/cnx#343 for details on why this has been implemented.

This has been implemented as a static build because looking up this info on ever deploy would be error-prone. The idea is to generate it once, check it in, and use it multiple times.

The REX host portion of this is configuration so that it can point at various environment instances of this service.

### DIFF
--- a/.update/do.py
+++ b/.update/do.py
@@ -1,0 +1,131 @@
+import click
+import requests
+from pathlib import Path
+
+from cnxcommon import ident_hash
+
+
+here = Path(__file__).parent
+CNX_HOST = 'archive.cnx.org'
+MAP_FILEPATH = here.resolve().parent / 'roles/webview/files/etc/nginx/uri-maps/rex-uris.map'
+
+
+def get_rex_release_json_url(host):
+    return f'https://{host}/rex/release.json'
+
+
+def get_book_slug(book_id):
+    url = (
+        "https://openstax.org/apps/cms/api/v2/pages/"
+        f"?type=books.Book&fields=cnx_id&format=json&cnx_id={book_id}"
+    )
+    book = requests.get(url).json()['items'][0]
+    return book['meta']['slug']
+
+
+def flatten_tree(tree):
+    """Flatten a tree to a linear sequence of values."""
+    yield dict([
+        (k, v)
+        for k, v in tree.items()
+        if k != 'contents'
+    ])
+    if 'contents' in tree:
+        for x in tree['contents']:
+            for y in flatten_tree(x):
+                yield y
+
+
+def rex_uri(book, page):
+    if page is None:
+        uri = f'/books/{book}'
+    else:
+        uri = f'/books/{book}/pages/{page}'
+    return uri
+
+
+def cnx_uri_regex(book, page):
+    if page is None:
+        uri_regex = f"/contents/({book['id']}|{book['short_id']})(@[.\d]+)?(/[-%\w\d]+)?$"
+    else:
+        uri_regex = f"/contents/({book['id']}|{book['short_id']})(@[.\d]+)?:({page['id']}|{page['short_id']})(@[\d]+)?(/[-%\w\d]+)?$"
+    return uri_regex
+
+
+def expand_tree_node(node):
+    result = {
+        'slug': node['slug'],
+        'title': node['title'],
+    }
+    result['id'], result['version'] = ident_hash.split_ident_hash(node['id'])
+    try:
+        # We raise an error for this... It maybe makes sense for the application of it in archive?
+        ident_hash.split_ident_hash(node['shortId'])
+    except ident_hash.IdentHashShortId as exc:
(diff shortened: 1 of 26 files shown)


.update/do.py
```diff
@@ -0,0 +1,130 @@
+import click
+import requests
+from pathlib import Path
+
+from cnxcommon import ident_hash
+
+
+here = Path(__file__).parent
+CNX_HOST = 'archive-staging.cnx.org'
```
BryanEliDimas: Is this same as or different than [rex-host](https://github.com/openstax/cnx-deploy/pull/1102/files#diff-fdbe56b1ec594bb5ca81b36486001878R106) /`rex_host`? And, it just seems to me like `CNX_HOST` will always point to staging?

.update/do.py
```diff
@@ -0,0 +1,130 @@
...
+MAP_FILEPATH = here.resolve().parent / 'roles/webview/files/etc/nginx/uri-maps/rex-uris.map'
+
+
+def get_rex_release_json_url(host):
+    return f'https://{host}/rex/release.json'
+
+
+def get_book_slug(book_id):
+    url = (
```
BryanEliDimas: This inline var seems dangerous—like it could change and silently fail; by "it" I mean the URL's accepted params or the path.

README.md
```diff
@@ -30,6 +30,15 @@ ansible-playbook -i environments/$ENV/inventory main.yml
 
 Please see the environment's README for specific details.
 
+## Updating generated content
+
+This project has generatated content in it. By running the `python3 .update/do.py` you are updating that content. It's your responsibilty to commit, push and create a pull request for the changes.
+
+```sh
+pip3 install --upgrade -r .update/requirements.txt
+python3 .update/do.py update-rex-redirects
```
BryanEliDimas: Doesn't it also need the `rex-host` arg?

roles/webview/templates/etc/nginx/sites-available/webview
```diff
@@ -7,76 +11,16 @@ server {
...
+        return 302 https://{{ rex_domain }}$rex_uri;
+    }
+
     location /ping {
         add_header Content-Length 0;
         add_header Content-Type text/plain;
         return 200;
     }
 
-{% if groups.authoring|length > 0 %}
```
BryanEliDimas: It isn't clear to me why this and all lines below this are getting deleted in this PR...

.update/do.py
pumazi: The rex hostname is different than the cnx hostname. However, this should be pointing at production rather than staging. It's just that the openstax/cnx#348 is not on production yet.

.update/do.py
pumazi: IMHO, all of this code is short-lived. So it's scrap that we are pulling together to build scaffolding for the larger more important project. This url is only now changing. Technically, it's a versioned API, so therefore it shouldn't change.

README.md
pumazi: Sure does, but we don't know what that is yet. Note, this PR is in draft state.

pumazi:
While reviewing this keep a few things in mind. The included 💀 commit will not be merged as part of this. This is being done like this because we aren't ready to actually redirect those urls. When we are ready, we'll run the provided command to generate this list. That will then get included in a tagged version of cnx-deploy.

README.md
pumazi: I'll put a placeholder in for now...

roles/webview/templates/etc/nginx/sites-available/webview
pumazi: It's cleanup. cnx-deploy is a "rebase and merge" repo. So the PRs here are a little looser since it's really the commits that are being merged rather than the whole PR.

pumazi: resolved in #1103

pumazi:
A modification needed to ensure the generic book URLs redirect to the first page of the book.

.update/do.py
```diff
@@ -0,0 +1,131 @@
...
+        "https://openstax.org/apps/cms/api/v2/pages/"
```
therealmarv: Very minor thing, should not it be `f"https://openstax.org/apps/cms/api/v2/pages/"` too?

.update/do.py
therealmarv: I agree with @pumazi comment. Because it is shortlived it's not a super big issue. I mean you can make url params super save but this should be only necessary for very dynamic content or when an id can be modified from outside (attack vector).

.update/do.py
pumazi: It doesn't actually have any formatting in it. It'd just be running the string formatting operation over it for no reason. But I could change it if you'd like.

therealmarv: It's really a non issue.

PR #1103, opened 2019-05-28
Remove authoring from webview nginx conf

pumazi:
Pretty straightforward as the PR title suggests.


BryanEliDimas:
@pumazi I was going to add a comment to let Bey and Britt know that this is nothing major and not connected to an issue but I think we could create a label instead. I don't know what to name it, feel free to do the honors 😄

PR #1104, opened 2019-05-28
 Provide a means of producing CNX urls for REX books 

pumazi:
This is to be used to produce all the variations of the CNX URLs for REX books.
It is primarily to be used for producing data for the load and functional tests in cnx-automation.

Addresses openstax/cnx#387

### DIFF
--- a/.update/do.py
+++ b/.update/do.py
@@ -119,12 +119,62 @@ def update_rex_redirects(rex_host):
             write_nginx_map(book_uri_map, out=fb)
 
 
+def generate_cnx_uris(book_id):
+    """\
+    Generates a list of URIs for a cnx book. The URIs are several variations
+    of the same page. This includes URIs with and without versions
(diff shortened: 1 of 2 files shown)


brittweinstein:
@pumazi what issue should this be attached to?

BryanEliDimas:
@brittweinstein @pumazi This PR seems like a duplicate of https://github.com/openstax/cnx-deploy/pull/1102

pumazi:
This is not a duplicate. It depends on commits from other PRs. It's not ready for review because it depends on other changes.

pumazi:
Note, I did include the issue this is in relation to in both the description and the branch name. 😛

pumazi:
The URLs this generates use generic numbers because we don't actually match on the numbers. And in fact, we don't want to match on specific numbers because we want all versions of the book to get redirected to REX.

PR #1138, opened 2019-06-18
Adding daily cronjob template to prune files

scb6:


templates/etc/cron.daily/var_www_files_cleanup
```diff
@@ -0,0 +1,6 @@
+{% for nfs_mount in nfs_mounts|default(default_nfs_mounts) %}
+find {{ nfs_mount.name }} -type f -name 'm*.zip' -mtime +21 -exec rm {} \;
```
DarkPrinceFrost: I'd personally change the commands to
`find -O3 {{ nfs_mount.name }} -mindepth 1 \( -name m\*.zip -type f -mtime +21 \) -print0 | xargs -0 rm -f`

The `-O3` optimizes the queries to find by name before statting any files. Also the `print0` and `xargs` are recommended just in case of strange file names

scb6: Thats fine. Updating.

PR #1143, opened 2019-07-01
More fixes for the vm environment

karenc:
- Only do "rsyslog : copy CA chain for graylog" if graylog is defined

  There's an encrypted file involved which is not necessary if graylog
  server or port is not defined (e.g. the `vm` environment).

- Define accounts_stub_users in vm host_vars

  This way we can override what's in group_vars/all/ and not use the
  encrypted vault_accounts_stub_users.

- Direct test for when database should be initialized

  Previously we do an indirect test of "if /etc/cnx and
  /etc/cnx/publishing exist", then the database should be initialized as
  well.  It's not robust because the playbook might have failed half way
  creating `/etc/cnx` but not initializing the database.
  
  Use "dbmigrator list" as the test.  It should output a warning message
  to say that the database hasn't been initialized (for migrations at
  least).

### DIFF
--- a/roles/publishing_common/tasks/main.yml
+++ b/roles/publishing_common/tasks/main.yml
@@ -121,6 +121,16 @@
   with_items:
     - cnx
     - cnx/publishing
+
+- name: check whether repository database is initialized
+  register: repo_db
+  become: yes
+  command: "/var/cnx/venvs/publishing/bin/dbmigrator --config /etc/cnx/publishing/app.ini --context cnx-db list"
+  ignore_errors: True
+
+- name: initialize repository database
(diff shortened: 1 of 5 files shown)


roles/publishing_common/tasks/main.yml
```diff
@@ -121,6 +121,16 @@
...
     - cnx
     - cnx/publishing
+
+- name: check whether repository database is initialized
+  register: repo_db
+  become: yes
+  command: "/var/cnx/venvs/publishing/bin/dbmigrator --config /etc/cnx/publishing/app.ini --context cnx-db list"
+  ignore_errors: True
+
+- name: initialize repository database
```
pumazi: It's more about initializing db-migrator, but it's close enough.

PR #1156, opened 2019-07-25
Update REX URI redirects from the latest rex/release.json

pumazi:
This updates the REX URI redirects using the slugs from qa.cnx.org to redirect to the canonical REX testing location at `rex-web.herokuapp.com`.

This also corrects the URIs for openstax/cnx#595, where the chapter number was missing on the review type pages.

---

This was generated using something like:

```sh
cd .update
python -m venv env
source env/bin/activate
pip install -r requirements.txt
python .update/do.py update-rex-redirects rex-web.herokuapp.com
```


pumazi:
I compared the offending URIs (lines under /books/intermediate-algebra/pages/1-exercises) with this and can confirm that the numbers are now on the pages.

PR #1165, opened 2019-08-08
Change content03 to point to exercises-qa, instead of prod

scb6:


scb6:
The `exercises-token` is blank in the vault so there is no change needed for that.

PR #1170, opened 2019-08-19
Move rex-uris.map from webview roles to different environments

karenc:
- Separate staging and prod files/ (symlink each file)

  We had `environments/staging/files/` and `environments/prod/files/`
  pointing to `environments/__prod_envs/files/`.  This is going to change
  because staging needs to redirect different books based on
  staging.openstax.org while prod needs openstax.org. (rex-uris.map is
  going to be different for staging and prod)

- Move rex-uris.map from webview roles to different environments

  We are going to generate different uris for rex redirects based on the
  rex domain for each environment so we need to have one rex-uris.map per
  environment.
  
  Right now all the rex-uris.map in different environments are the same,
  but they are going to regenerated using different rex sites.

- Set mode from 0644 to 0755 for /etc/nginx/uri-maps

  0644 = readable and writable for owner, readable for group and all
  
  0755 = readable, writable and executable for owner, readable and
  executable for group and all
  
  For directories, the readable flag usually comes with the executable
  flag because listing directories seems to require permission to execute
  them.  Without the executable flag:
  
  ```
  karen@qa00:~$ ls -l /etc/nginx/uri-maps
  ls: cannot access '/etc/nginx/uri-maps/rex-uris.map': Permission denied
  ls: cannot access '/etc/nginx/uri-maps/rex-uris.map.bak': Permission denied
  total 0
  -????????? ? ? ? ?            ? rex-uris.map
  -????????? ? ? ? ?            ? rex-uris.map.bak
  ```


karenc:
@pumazi `environments/prod/files/*-requirements.txt` and `environments/staging/files/*-requirements.txt` are symlinks pointing to `environments/__prod_envs/files/*-requirements.txt`.  So they should be in sync.

PR #1172, opened 2019-08-20
Enable rex redirects using staging.openstax.org for staging

karenc:
- Enable rex redirects using staging.openstax.org for staging

  Now that we have separate rex-uris.map files for each environment, we
  can safely use different rex domains for different environments.

- .update/do.py generate rex-uris.map using cnx-deploy environments

  `.update/do.py` used to update a specific file
  `roles/webview/files/etc/nginx/uri-maps/rex-uris.map`.  We now have
  different rex-uris.map for each environment in cnx-deploy, so
  `.update/do.py` needs to take a cnx-deploy environment as argument
  (instead of rex host), finds the `rex_domain` in the vars file and
  generates the rex-uris.map in the correct directory.

- Regenerate rex-uris.map for staging (release id: master/84f7661)


karenc:
The books that will redirect to staging.openstax.org are:

```
anatomy-and-physiology
biology-2e
biology-ap-courses
chemistry-2e
chemistry-atoms-first-2e
concepts-biology
```

PR #1173, opened 2019-08-21
Allow collection version as page version in regexes used in REX redirects

jpslav:
@kjdav noticed that CNX -> REX redirects were not working for end of chapter pages and end of book pages (key terms, etc).  She also pointed out to me that these pages are different in that their page version is actually the collection version.

So when a user went to Chem 2e's chapter 1 key terms:

https://staging.cnx.org/contents/f8zJz5tx@8.1:1bA1DyjO@8.1 

The landed on CNX instead of REX.

I think this is because the regex used in the nginx mapping file only allows for page versions to be digits (`[\d]+`).  This PR changes that part of the regex to allow dots as well (`[.\d]+`) just as is done earlier in the regex for collection versions.

The README says that when I make this change I need to regenerate the mapping files that are committed in this repo, but I leave that to more capable CE hands :-)


karenc:
**Bug Verified**: [Key Terms for Chapter 1 of Anatomy & Physiology](https://archive-staging.cnx.org/contents/14fb4ad7-39a1-4eee-ab6e-3ef2482e3e22@16.1:d66bf2e4-2231-5a2b-a75e-25fc24b34b35@16.1.html) doesn't redirect to staging.openstax.org, but [1.7 Medical Imaging](https://staging.cnx.org/contents/14fb4ad7-39a1-4eee-ab6e-3ef2482e3e22@16.1:bda09c0d-e898-4653-9d37-3cb49fab55e5@8) does.

karenc:
Fix verified locally. :heavy_check_mark: I just need to update the `rex-uris.map`s.

PR #1174, opened 2019-08-24
Change dev envs to use staging.openstax.org as rex_domain

karenc:
The `rex_domain` setting is used for generating rex redirects using
cnx-rex-redirects script.

rex-web.herokuapp.com doesn't have /rex/environment.json, causing
cnx-rex-redirects to fail.  We are going to just use
staging.openstax.org for all non-prod environments.


karenc:
We talked about it in the card check in, and it's ok to change all non-prod environments to use staging.openstax.org

PR #1192, opened 2019-09-25
Update rex redirects for prod

karenc:
+astronomy
+college-physics
+college-physics-ap-courses
+microbiology


karenc:
`environments/prod/files/etc/nginx/uri-maps/rex-uris.map` and `environments/staging/files/etc/nginx/uri-maps/rex-uris.map` are exactly the same, so I think it's ok to merge this.

PR #1222, opened 2019-11-20
Update rex redirects for __dev_envs

karenc:
-astronomy
+astronomy
-astronomy
+astronomy

### DIFF
--- a/environments/__prod_envs/files/publishing-requirements.txt
+++ b/environments/__prod_envs/files/publishing-requirements.txt
@@ -12,7 +12,7 @@ cnx-easybake==1.2.7
 cnx-epub==0.20.1
 cnx-publishing==0.17.1
 cnx-query-grammar==0.2.2
-cnx-recipes==1.42.0
+cnx-recipes==1.43.0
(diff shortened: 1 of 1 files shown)


karenc:
@helenemccarron No, katalyst01 has rex redirects disabled: 

https://github.com/openstax/cnx-deploy/blob/master/environments/katalyst01/group_vars/all/vars.yml#L72

I guess we can remove the file from `environments/katalyst01`

In any case, we shouldn't merge this PR because the page slug has changed on staging but it doesn't look right for production.

helenemccarron:
Ok, thank you for the clarification. From a recipes perspective, at this point, we are requesting that Katalyst01 and easyvm5 do *not* redirect so we can continue functional testing on webview. cc @openstaxalina

karenc:
@helenemccarron No, this PR doesn't touch that variable so katalyst01 still doesn't redirect to rex.

karenc:
Ok great!  I've changed the concourse pipeline to use prod archive instead of staging archive and now there are no changes :tada:

PR #1278, opened 2020-03-24
Modify webview REX redirect map block to use $uri

rnathuji:
It was found that redirects did not get mapped as expected when the
request URIs included query parameters. This change is intended to
address this issue by using the $uri variable instead of $request_uri
in the map block as it does not include the query string which we
don't need / use.


rnathuji:
Relevant [comment](https://github.com/openstax/cnx/issues/921#issuecomment-603273413) in parent issue cnx#921 for this PR

PR #1292, opened 2020-04-01
added a concourse pipeline to auto-deploy qa on new tags

m1yag1:


m1yag1:
Probably not going to finish this as it's probably not going to get done.

PR #1296, opened 2020-04-06
Improve VM environment docs and Vagrantfile

rnathuji:
Adding some structure and content to better articulate the options
for using the VM environment. Also includes minor cleanup items
in the Vagrantfile that came up during documentation updates.


omehes:
looks good and works as intended

PR #1311, opened 2020-04-27
updated dependencies to reflect latest changes in cnx-archive 4.14.2

m1yag1:


m1yag1:
Not needed as pyup PR should account for these. I initially wanted to ignore pyup but seems to work ok if all applications are updated at the same time.

PR #1342, opened 2020-06-25
Update Vagrant deployment for rhaptos.org firewall

rnathuji:
External access to dist.rhaptos.org will be blocked by a firewall for
security reasons. In Vagrant deployments, this will prevent packages
from being pulled during buildout in the zope role causing failures.
As a work around, we'll inject an entry in the cnx-target `/etc/hosts`
for dist.rhaptos.org that points to the Vagrant host where users can
setup an ssh tunnel to bastion2.cnx.org which will continue to have
access to dist.rhaptos.org.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,10 +52,13 @@ Vagrant.configure("2") do |config|
       fi
     SHELL
 
-    # Run playbook to setup /etc/hosts on target host
+    # Run vagrant specific playbook for preliminary setup
     host.vm.provision "ansible_local" do |ansible|
-      ansible.playbook = "configure_hosts.yml"
+      ansible.playbook = "configure_vagrant.yml"
       ansible.inventory_path = "environments/vm/inventory"
+      ansible.extra_vars = {
+        rhaptos_tunnel_ip: "10.0.10.1"
+      }
       ansible.limit = "all"
       ansible.verbose = "-vvv"
     end
--- a/configure_vagrant.yml
+++ b/configure_vagrant.yml
@@ -1,0 +1,24 @@
+---
+# Playbook for configuration specific to Vagrant environment
+
+- import_playbook: configure_hosts.yml
+
+# NOTE: This is to help address firewall settings that prevent external access
+# to dist.rhaptos.org. It expects an ssh tunnel on the Vagrant host to be
+# running during provisioning. (See details in environments/vm/README.md)
+- name: "configure dist.rhaptos.org in /etc/hosts"
+  hosts:
+    - zope
+  tasks:
+    - name: remove old rhaptos.org entries from target /etc/hosts
+      become: yes
+      lineinfile:
+        dest: "/etc/hosts"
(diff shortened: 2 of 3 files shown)


Vagrantfile
```diff
@@ -52,10 +52,13 @@ Vagrant.configure("2") do |config|
       fi
     SHELL
 
-    # Run playbook to setup /etc/hosts on target host
+    # Run vagrant specific playbook for preliminary setup
     host.vm.provision "ansible_local" do |ansible|
-      ansible.playbook = "configure_hosts.yml"
+      ansible.playbook = "configure_vagrant.yml"
```
tomjw64, line 57: it seems this is a new file. Is `configure_hosts.yml` still used? I can't find any other usage with a quick GitHub search/

Vagrantfile
```diff
@@ -52,10 +52,13 @@ Vagrant.configure("2") do |config|
...
       ansible.inventory_path = "environments/vm/inventory"
+      ansible.extra_vars = {
+        rhaptos_tunnel_ip: "10.0.10.1"
```
tomjw64, line 60: Interesting. Why this address? (This is more curiosity from lack of knowledge than actual review)

rnathuji: Good question -- so this relying on the private_network subnet defined (10.0.10/24) where the gateway IP of `.1` is (by convention) associated with the host's interface on the network.

Vagrantfile
m1yag1: The original instructions before vagrant used this file to configure the /etc/hosts file. I imagine we would still need this file if we were to run the playbook outside of vagrant.
https://github.com/openstax/cnx-deploy/pull/1342/files#diff-ba2b72f148fb98bb8e8549f4336d2893R18-R23

Vagrantfile
tomjw64: ah okay. so this is telling the deployment there is some other machine at 10.0.10.1 where it can download `dist.rhaptos.org` packages? And we use the VM host here because that's where we have the ability to tunnel through bastion2? I'm surprised that the `10.0.10.1` on the VM is the same as the VM host, on the other hand. For my sake, can you point out how the VMs and the host are connected via the private network?

rnathuji: Sure -- so the private network breaks down as:

* 10.0.10.1 - Your host machine (the address of its interface on the private network)
* 10.0.10.2 - The `cnx-deploy` VM (the IP for the VM which executes the deployment via ansible)
* 10.0.10.10 - The `cnx-target` VM (the IP for the VM which runs the cnx stack)

Per your comment, the insertion of the entry in `/etc/hosts` in `cnx-target` will cause it to resolve `dist.rhaptos.org` to `10.0.10.1` (your host) where the tunnel runs and forwards the request.

LMK if that helps / answers your question.

tomjw64: Thanks for the explanation!

PR #1364, opened 2020-08-19
Archive depend fix

m1yag1:
In order to fix the dependency issue I've created a new virtualenv where
archive is installed as a package with pip. A separate task will install any
outside dependencies for deployment so that these are more explicit. I
suspect there are some defined in archive-requirements.txt that are not
needed. To facilitate a possible migration to python 3 I've created a
folder in the venv dir for python2 virtualenvs. When we move over we can
remove the old archive venv and leave room open to create a python3 
folder if needed.

* Added task to create a new python2 folder in the venv folder
* Added a task to install archive as a package using pip instead of using the archive-requirements.txt. This allows archive 
  package to control its own dependencies.
* Added a task to ensure setuptools for the new virtualenv is using
  setuptools<45 which causes errors when installing packages in python2
  land.

### DIFF
--- a/environments/__prod_envs/vars/versions.yml
+++ b/environments/__prod_envs/vars/versions.yml
@@ -1,2 +1,3 @@
 webview_version: v0.54.0
 press_version: v8.1.7
+archive_version: 4.15.9
--- a/roles/archive/templates/etc/supervisor/conf.d/archive.conf
+++ b/roles/archive/templates/etc/supervisor/conf.d/archive.conf
@@ -1,6 +1,6 @@
 {% for i in range(0, hostvars[inventory_hostname].archive_count|default(1), 1) %}
 [program:archive{{ i }}]
-command=/var/cnx/venvs/archive/bin/pserve --server-name=instance{{ i }} /etc/cnx/archive/app.ini
+command={{ archive_pserve_path }} --server-name=instance{{ i }} /etc/cnx/archive/app.ini
 
 {% endfor %}
 
--- a/update_versions.yml
+++ b/update_versions.yml
@@ -9,13 +9,13 @@
 - name: get archive versions
   hosts: frontend
   tasks:
-    - shell: /var/cnx/venvs/archive/bin/pip freeze | grep 'cnx-archive'
+    - shell: "$(dirname {{ archive_pserve_path }})/pip freeze | grep 'cnx-archive'"
       register: archive_version
       when: groups.archive
       delegate_to: "{{ item }}"
       with_items:
         - "{{ groups.archive|first }}"
-    - shell: /var/cnx/venvs/archive/bin/pip freeze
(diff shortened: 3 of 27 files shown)


roles/archive/templates/etc/supervisor/conf.d/archive.conf
```diff
@@ -1,6 +1,6 @@
 {% for i in range(0, hostvars[inventory_hostname].archive_count|default(1), 1) %}
 [program:archive{{ i }}]
-command=/var/cnx/venvs/archive/bin/pserve --server-name=instance{{ i }} /etc/cnx/archive/app.ini
+command={{ archive_pserve_path }} --server-name=instance{{ i }} /etc/cnx/archive/app.ini
```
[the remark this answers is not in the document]
m1yag1: Yes, this would need to happen. I figured it could be added after being tested on QA.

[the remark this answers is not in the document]
rnathuji: Cool -- Mostly I wasn't sure if this would break deployments to other environments when merged, but it sounds like that'll be addressed beforehand.

environments/__prod_envs/vars/versions.yml
```diff
@@ -1,2 +1,3 @@
 webview_version: v0.54.0
 press_version: v8.1.7
+archive_version: 4.15.8
```
rnathuji, line 3: Should this reflect the latest version update to archive? (It's currently at `4.15.9`)

update_versions.yml
```diff
@@ -9,13 +9,13 @@
 - name: get archive versions
   hosts: frontend
   tasks:
-    - shell: /var/cnx/venvs/archive/bin/pip freeze | grep 'cnx-archive'
+    - shell: /var/cnx/venvs/python2/archive/bin/pip freeze | grep 'cnx-archive'
```
rnathuji, line 12: During the transition period, is it possible for this to report the wrong version if (for whatever reason) we have to revert some environment to use `/var/cnx/venvs/archive/bin/pserve` and (again, for whatever reason) the version in `environments/__prod_envs/vars/versions.yml` doesn't match `environments/__prod_envs/files/archive-requirements.txt`? 

If so, is it possible to somehow base the pip on `{{ archive_pserve_path }}`?

environments/__prod_envs/vars/versions.yml
m1yag1: Yes, we should update to the latest version

update_versions.yml
m1yag1: I'm not removing the old virtualenv so I don't think the scenario you are describing would happen. If a revert did occur  it would go back to the old virtualenv, no? I could be missing something.

Maybe we should have a quick call?

rnathuji (approved):

PR #1377, opened 2020-09-16
Update zope role and vm environment for new dist.cnx.org

rnathuji:


m1yag1 (approved):
I think it could be more clear that the vm_vars.yml should go into the root project directory but other than that it worked for me.

PR #1385, opened 2020-09-29
backout archive-requirements.txt tasks from archive role

m1yag1:
This is rebased from another branch. MERGE [#1377](https://github.com/openstax/cnx-deploy/pull/1377) FIRST and rebase again!

This backs out tasks in the archive role that install dependencies via the archive-dependencies.txt file in the archive role. We moved this over to being installed as a package and so this old code needs to be removed.

### DIFF
--- a/roles/archive/tasks/main.yml
+++ b/roles/archive/tasks/main.yml
@@ -40,9 +40,8 @@
     mode: 0755
     owner: "{{ venvs_owner|default('www-data') }}"
 
-- name: create a new folder for python2 virtualenv 
+- name: create a folder for python2 virtualenvs
   become: yes
-  when: not venvs_dir.stat.exists
   file:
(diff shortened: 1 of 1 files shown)


rnathuji:
From a brief look, I guess while we can remove the use of `archive-requirements.txt` from the archive role, it's still needed in a few places for the time being (e.g. I see a reference in the `cnx_database` role). Is that correct?

Also, I merged [PR 1377](https://github.com/openstax/cnx-deploy/pull/1377) so you should be good to rebase + merge this.

rnathuji (approved):

m1yag1:
@rnathuji , yes it's still used in a couple of other areas. If you look in the issue I identified another area where it's used. The entire virtualenv gets created by the plpy_imports task 🤦 .

PR #1393, opened 2020-10-14

m1yag1:
-algebra-and-trigonometry
+algebra-and-trigonometry
-precalculus
+precalculus
-precalculus
+precalculus


c0ldcalamari (approved):

omehes:
Verified both books on latest staging deploy. Both redirects tests passed. looks good

PR #1407, opened 2020-12-01
Follow up work for #1133: remove publishing-requirements.txt from the repo entirely

m1yag1:


environments/__prod_envs/files/publishing-requirements.txt
```diff
@@ -1,64 +0,0 @@
...
-certifi==2020.11.8
-chardet==3.0.4
-cnx-archive==4.15.14
-cnx-common==1.3.6
-cnx-db==3.5.3
-cnx-easybake==1.2.7
-cnx-epub==0.25.2
-cnx-publishing==0.17.23
-cnx-query-grammar==0.2.2
-cnx-recipes==1.66.0
```
rnathuji, line 15: I noticed that in `cnx-publishing` this dependency is [not pinned](https://github.com/openstax/cnx-publishing/blob/bb29d52af59b9a6f00cdb434e29f7935d354e44a/requirements/main.txt#L14). Should we set it explicitly via `==` there as part of this? 

Also, we can let styles team folks who update `cnx-recipes` versions know that the updates should flow through `cnx-publishing` repo moving forward.

tomjw64 (approved):

environments/__prod_envs/files/publishing-requirements.txt
m1yag1: Oh wow, that's interesting. I'd like to pin it explicitly to a version. I'm trying to think of the rationale of why this was pinned this way in the first place. I guess they wanted to use latest of cnx-recipes when doing development but pinning to an exact version to deployment. In that case, yea we should update this and let the styles team know.

rnathuji: Yeah, that's probably what happened. It looks like @tomjw64 and @therealmarv are going to pin the version in their [updates today](https://openstax.slack.com/archives/C0LA54Q5C/p1606930001434000?thread_ts=1606928568.430600&cid=C0LA54Q5C), so once that's done we can let the styles team know.

tomjw64: Oh, I've actually just reminded myself why we pinned it in this manner. I think that the qa server for ce-styles also used cnx-deploy, and so without having a slightly open pin, they were unable to test recipes in development. Is there a way to prevent them from experiencing that problem again even with a stronger pin?

m1yag1: i don't see how they would be able to test recipes in development since something would still need to be tagged.

PR #1409, opened 2020-12-03
Pin cnx-recipes version in publishing environment

tomjw64:


environments/__prod_envs/vars/versions.yml
```diff
@@ -10,3 +10,4 @@ archive_deploy_dependencies:
 
 publishing_deploy_dependencies:
   - { "db-migrator": 1.1.0 }
+  - { "cnx-recipes": 1.65.0 }
```
rnathuji, line 13: Oh, I think we want to incorporate the update to version `1.66.0` here

rnathuji (approved):

PR #1464, opened 2021-11-08

staxbot:
-concepts-biology
+concepts-biology
-física-universitaria-volumen-1;
+física-universitaria-volumen-1


m1yag1 (approved):

omehes:
tested: looks good, apart from chapter 7.3 which seems to have different title (unless, this was the fix and is not deployed):
https://staging.cnx.org/contents/b3c1e1d2-839c-42b0-a314-e119a8aafbdd:6-3MVU-j/7-3-variations-in-meiosis

PR #1466, opened 2022-10-25

staxbot:
-principles-macroeconomics
+principles-macroeconomics
-principles-microeconomics
+principles-microeconomics
+fizyka-dla-szkół-wyższych-tom-1
+fizyka-dla-szkół-wyższych-tom-2
+fizyka-dla-szkół-wyższych-tom-3


philschatz (approved):
The micro/macro redirects are correct (the old destination in the diff is broken while the new destination works) and the fizyka destinations all work

